PR #1, opened 2018-10-27
CSS: StylePreprocessor, css color update suggestion

lu-c:
1. Added stylePreprocessorOptions to angular.json so files do not need to know relative path to style files.
2. Moved styles from src/app/theme to src/assets/styles
3. Did some preliminary work in updating fonts to use a different color schema. The color mappings may not be perfect right now, this was a tentative start.


reblace:
One of the reasons I kept the core theme stuff out of the src/assets directory is because that directory is directly copied into the dist dir when you build. It is meant more for things you want to have served as-is and not get ingested and processed by the assets pipeline. So, that's not really what you want to do with the scss files here.

Maybe we could go with src/styles or src/app/styles?

lu-c:
Alright, I understand, didn't realize it had that effect on build. I moved everything to src/styles instead

reblace (approved):

PR #2, opened 2018-10-27
Upgrade to ng7

reblace:
Upgrade to ng7 and some other fixes. Also adding some site implementation examples. This PR fixes the Firefox AOT build issue.

I had to revert the stylePreprocessorOptions change:
- It appears that in ng7, imports of partials doesn't work without including the leading underscore.
- I'm also on the fence about using automatic path lookups when they are really well established patterns just cause it makes it less apparent where a file really is. Plus, the IDE would need to know how to read those options from angular.json, otherwise you get a big red error about path not found.


brianghig:
@reblace it looks like the change for the `config()` method to return the promise is actually causing the web app to fail to start when the `proxy.conf.json` mapping for `/api` is not available. Should we assume that the `ngx-starter` will never run without the corresponding `/api/config` REST endpoint available?

Steps to reproduce:
```
npm start
```

Output:
```
...
ℹ ｢wdm｣: Compiled successfully.
[HPM] GET /api/config -> http://localhost:3000
[HPM] Error occurred while trying to proxy request /api/config from localhost:4200 to http://localhost:3000 (ECONNREFUSED) (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

reblace:
That wasn't the intent. I think we just need to complete the behavior subject even if the call to get the config fails.

For now, I want the app to wait till the config is loaded to initialize. If it fails, the app will load but be have default options. Eventually, we probably want the app to go into an "Initializing" state or something and have it show an error screen if the server is unavailable.

PR #10, opened 2018-11-02
migrating help feature

jrassa:
Bringing over help feature from existing app, updated for ng7.  Also include breadcrumb component.


brianghig:
@jrassa functionality looks good! Thanks for bringing this over. Are there any SCSS files for the breadcrumbs that we can add? At the moment, it looks like it may be in need of padding, text decoration, and hover indicators.

screen shot 2018-11-05 at 2 06 27 pm

jrassa:
@brianghig styling has been updated.  One question I had is that I wasn't sure if help should go under common or core.  I went with common so that it could be excluded more easily if an app didn't want/need it, but I would be fine with either.

reblace:
Probably core. Core is for things that are a part of the site or are only
expected to be created once per site. Common is for reusable utilities that
are used by a lot of components.

brianghig:
Code inspection looks good, but I haven't executed the latest linting and tests on this branch.

Issue #12, opened 2018-11-06
Add User Management for Admins


brianghig:
In order to allow administrators to manage users, add a user management module that allows for administrators to CRUD all aspects of a user account, including passwords, roles, metadata, etc.

brianghig:
@jrassa @reblace would you see this administrative-facing functionality baked in as a core module in the `ngx-starter`, or as an entirely separate application that is focused only on administrative functionality?

My preference is to include it as a `core` module, but could see an argument for a separate app.

PR #15, opened 2018-11-07
Admin User Listing and Management #12

brianghig:


brianghig:
@reblace @jrassa @ymao1 this is still very much a work in progress, but I was hoping to get your thoughts on the common "pageable" module that would encapsulate components like paging options, sorting, pageable-table, etc. that I'm porting in to this branch. Have you found other / better means of doing this, or preferred third party options for tables?

jrassa:
@brianghig did you base this on the mean2-starter?  There have been enhancements in downstream apps that I guess never made it there.

Also, I can't speak for the original intent., but the Pager and sort models could have uses outside of the pageable table.  I presume that is why they weren't included with it originally.

ymao1:
We do use the pager component outside of the context of the pageable table but I think it's reasonable to put it inside a 'pageable' module. I know on our project we are still making use of the pageable table vs another dependency.

brianghig:
@jrassa @ymao1 thanks! Yes, this was based on `mean2-starter` and I did want to continue the separation of `pager` and `sort` models from the `pageable-table`. At the moment, it's all grouped under / exported from a `pageable` module whose separate components can be imported as-needed.

reblace:
I was going through and commenting on it, but I think I want to have a conversation about this PR.

I'm hesitant to put things that are overly opinionated about style and structure into common. They should go into the site section. The example site can be used as a reference app, but anyone who is going to create a custom look and feel will want to have reusable components that aren't as assertive about DOM structure. That would suggest that things like pipes and simple services can go in common, something like the paging table shouldn't. At least not the part of it that contains the template.

What I'm trying to avoid is a situation where someone wants to create a custom look and feel, but ends up with a bunch of stuff in common that they never use because it has a style they don't want to use. If someone wants to adopt the style of the reference application, they would just copy it over and use it. Does that make any sense? We can sync up when I get back to talk about it in more detail.

brianghig:
@reblace it seems like most of the components and modules in `common` are more appropriate to be published as their own, standalone projects to be imported ad-hoc.

Regarding the expansion of the built-in administrative capabilities, are you suggesting that those should not be included in this reference app at all? Or rather that they be part of a new `admin` module under `site` that can be included / excluded at will in `site.module.ts`? With the latter approach, removing that inclusion from the `site.module.ts` file would easily do the trick of eliminating unnecessary bundles.

brianghig:
@jrassa @reblace this PR is now updated to include the Admin Users List and Admin User Update page.

- All tests are passing
- All linting rules are passing
- Completed a manual test of an admin updating a user's name and password

As part of that testing, I noticed that a non-admin user is allowed to browse to the AuthGuard'ed `/admin/users` endpoint, but I think we can address that in a following ticket.

jrassa:
Looks good.

Issue #17, opened 2018-11-08
Remove import star aliasing whenever appropriate


brianghig:
This should help to reduce the overall size of the bundle, by eliminating unnecessary imports from vendor libraries.

For example, `import * as _ from 'lodash';` that only uses the `_.isEmpty` function can simply be `import { isEmpty } from 'lodash';`

brianghig:
Resolve with #21

PR #21, opened 2018-11-09
Targeted Imports and Prod Build Fixes

brianghig:
This addresses #17 

- Moves all * imports, where possible, to targeted imports
- Added `bundle-report` NPM script that will create the production distribution
- Updated Travis CI to run production build during execution
- Fixed issue in Admin List Users HTML that caused the production build to fail

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,11 +6,12 @@
     "start": "ng serve --proxy-config proxy.conf.json",
     "start:prod": "ng serve --proxy-config proxy.conf.json --prod",
     "build": "ng build",
-    "build:prod": "ng build --prod",
+    "build:prod": "ng build --prod --stats-json",
     "test": "ng test",
     "test:ci": "ng test --watch=false --code-coverage",
     "lint": "ng lint",
-    "e2e": "ng e2e"
+    "e2e": "ng e2e",
+    "bundle-report": "npm run build:prod && webpack-bundle-analyzer dist/ngx-starter/stats.json"
   },
   "private": true,
   "dependencies": {
@@ -53,6 +54,7 @@
     "protractor": "5.4",
     "ts-node": "7.0",
     "tslint": "5.11",
-    "typescript": "3.1"
+    "typescript": "3.1",
+    "webpack-bundle-analyzer": "3.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test:ci": "ng test --watch=false --code-coverage",
     "lint": "ng lint",
     "e2e": "ng e2e",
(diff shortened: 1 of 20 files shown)


brianghig:
@jrassa @reblace @ymao1 @lu-c the only "big" import we have left is `moment`, which is a dependency from the `AgoDatePipe` and `UtcDatePipe`. If we're okay with the size, then we can leave that alone. However, since `moment` doesn't allow for clean tree-shaking in its current state, we may want to consider moving to something like `date-fns` that has a smaller footprint (as long as it supports the capabilities we still need).

Do many other projects have extensive dependencies on `moment`, or is that falling out of favor?

jrassa:
@brianghig Did you verify that the lodash changes actually resulted in a smaller bundle? I come across [this article](https://medium.com/@armno/til-importing-lodash-into-angular-the-better-way-aacbeaa40473) earlier today that indicates that it will still pull in all of lodash unless we import like this:
```
import { isArray } from 'lodash/isArray';
import { keys } from 'lodash/keys';
import { toString } from 'lodash/toString';
```

brianghig:
@jrassa thanks for that! I had started putting together an analysis of the size reduction, but never finished to post it with this PR. That said...

Good call on your part of looking that up! I fixed the `bundle-report` script to give us granular details of each module, and it's showing that `lodash` is still pulling in the whole thing (as well as plenty of code from `ngx-bootstrap` that we might want to look at.

Looks like there's more to fix with this PR before we call it "done"

screen shot 2018-11-09 at 5 01 16 pm

brianghig:
@jrassa I just updated the lodash imports, and found that they required `import fn from 'lodash/fn';` instead of `import { fn } from 'lodash/fn';`. According to the output from `npm run bundle-report`, this had a nice reduction on the bundle size:

State | Stat Size | Parsed Size | Gzipped
--- | ---- | --- | ---
Before | 527 KB | 527 KB | 93 KB
After | 73 KB | 88 KB | 16 KB
Change | -86.1% | -83.3% | -82.8%

It looks like we have a lot of other opportunities to save even more space from the new reduced baseline of 5.75 MB vendor files.

package.json
```diff
@@ -10,7 +10,8 @@
     "test": "ng test",
     "test:ci": "ng test --watch=false --code-coverage",
     "lint": "ng lint",
-    "e2e": "ng e2e"
+    "e2e": "ng e2e",
+    "bundle-report": "ng build --stats-json && webpack-bundle-analyzer dist/ngx-starter/stats.json"
```
reblace: For consistency, should this script be called "build:bundle-report" or something like that?

brianghig: I like it. I moved it to `build:bundle-report` and added a section into the README that describes each of the available NPM scripts so folks don't have to hunt through the `package.json` file.

PR #24, opened 2018-11-12
implement feedback support. modal for submitting feedback.  admin pag…

jrassa:
…e for viewing feedback.

includes support for dynamically registering admin topics.  Dynamic registration populates navbar popover and tabs on admin pages.

### DIFF
--- a/src/app/core/export-config.service.ts
+++ b/src/app/core/export-config.service.ts
@@ -1,0 +1,21 @@
+import { Injectable } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
+import { Observable } from 'rxjs';
+
+@Injectable()
+/**
+ * Admin management of users
+ */
+export class ExportConfigService {
+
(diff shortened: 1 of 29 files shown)


src/app/core/export-config.service.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Admin management of users
+ */
+export class ExportConfigService {
+
+	constructor(private http: HttpClient) {}
+
+	postExportConfig(type: string, config: any): Observable<any> {
+		return this.http.post(
+			'/requestExport',
```
brianghig: @jrassa should this be `/api/requestExport` ?

brianghig (changes requested):

src/app/core/export-config.service.ts
jrassa: good catch

brianghig (approved):

Issue #25, opened 2018-11-13
Add styling to admin filters


brianghig:
The Admin Users and EUA listings have unstyled Column selectors and quick filters. Let's add a baseline styling that provides a consistent branding and UX design.

brianghig:
@jrassa @ymao1 do we have any default styling that we can bring over for these left-hand filters?

jrassa:
@brianghig I am working on updating some stuff with the filters now.  I will include better default styling.

brianghig:
Thanks @jrassa ! Feel free to assign this issue to yourself if you'd like. I'm working on #19 at the moment.

jrassa:
Resolved by #28

PR #28, opened 2018-11-14
common components

jrassa:
Adds quick-filter, quick-column-toggle, search-table, and system-alerts components to common.  Updates to admin components to use these.

### DIFF
--- a/src/app/core/feedback/admin/admin-list-feedback.component.html
+++ b/src/app/core/feedback/admin/admin-list-feedback.component.html
@@ -15,7 +15,7 @@
 				<div class="col-md-12 form-group">
 					<div class="btn-group pull-right dropdown">
 						<button type="button" class="btn btn-sm btn-outline-secondary" (click)="export()">
-							<i class="fa fa-download"></i> Export
+							<span class="fa fa-download"></span> Export
 						</button>
 					</div>
 				</div>
@@ -35,7 +35,7 @@
 					<td>{{ feedback?.type }}</td>
 					<td>{{ feedback?.body }}</td>
 					<td>{{ feedback?.url }}</td>
-					<td [tooltip]="feedback.created" container="body" placement="top">{{ feedback?.created | agoDate:false }}</td>
+					<td [tooltip]="feedback?.created" container="body" placement="top">{{ feedback?.created | agoDate:false }}</td>
 				</ng-template>
 
 				<ng-template named-template="empty-table">
--- a/src/app/core/feedback/admin/admin-list-feedback.component.html
+++ b/src/app/core/feedback/admin/admin-list-feedback.component.html
@@ -2,40 +2,51 @@
 
 	<h3 class="page-header">
 		System Feedback
-
-		<div class="header-btns pull-right">
-			<button type="button" class="btn btn-primary pull-right" (click)="export()">
-				<i class="fa fa-download"></i> Export
-			</button>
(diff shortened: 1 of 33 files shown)


jrassa:
This also addresses #25

src/styles/_display.scss
```diff
@@ -71,3 +71,18 @@
 		padding-left: 5px * $i;
 	}
 }
+
+//----------------------------------------
+// Positioning
+//----------------------------------------
+.pull-left {
```
brianghig, line 78: @jrassa do we need these, or are they dupes from what bootstrap provides?

src/app/core/feedback/admin/admin-list-feedback.component.html
```diff
@@ -2,40 +2,51 @@
...
-	</pageable-table>
+	<div class="row padding-top-5">
+		<div class="col-md-12">
+
+			<!-- Export -->
+			<div class="row">
+				<div class="col-md-12 form-group">
+					<div class="btn-group pull-right dropdown">
+						<button type="button" class="btn btn-sm btn-outline-secondary" (click)="export()">
+							<i class="fa fa-download"></i> Export
```
brianghig: Now that Font Awesome supports "fa" classes on "span" tags in order to be more semantically accurate, I think we should start moving to `<span>` vs. `<i>`. Sounds like a new ticket to me, though 😄

src/app/core/feedback/admin/admin-list-feedback.component.html
```diff
@@ -2,40 +2,51 @@
...
+					</th>
+				</ng-template>
+
+				<ng-template named-template="table-row" let-feedback>
+					<td>{{ feedback?.creator?.name }}</td>
+					<td>{{ feedback?.creator?.email }}</td>
+					<td>{{ feedback?.type }}</td>
+					<td>{{ feedback?.body }}</td>
+					<td>{{ feedback?.url }}</td>
+					<td [tooltip]="feedback.created" container="body" placement="top">{{ feedback?.created | agoDate:false }}</td>
```
brianghig: Should this `feedback` object be optional for consistency with the rest of the displays? Or does the difference of a data binding cover the null case?

brianghig (changes requested):
Only one minor comment about the `pull-left` and related CSS classes that might already be provided by Bootstrap.

Are there any tests available to cover this new functionality?

src/styles/_display.scss
jrassa: As far as I can tell, `pull-left` and `pull-right` aren't coming from bootstrap.  They are currently coming from font-awesome, but they are marked as deprecated and will be removed in v5.  It also just seemed improper to rely on a class from font-awesome for general styling.

src/app/core/feedback/admin/admin-list-feedback.component.html
jrassa: fixed

src/styles/_display.scss
reblace: https://getbootstrap.com/docs/4.0/migration/#utilities

Bootstrap v4 moved to "float-left" and "float-right".
http://getbootstrap.com/docs/4.0/utilities/float/

brianghig: Thanks @reblace. I added #30 to cover that migration.

PR #32, opened 2018-11-21
Auth Controls

brianghig:


src/app/core/auth/auth.guard.ts
```diff
@@ -46,19 +51,78 @@ export class AuthGuard implements CanActivate {
...
-			map((session) => {
-				// The user isn't authenticated
+			switchMap((session) => {
+				// The user isn't authenticated, try reloading
 				if (session == null) {
-					// Send them to signin with a redirect to the previous URL
-					this.sessionService.setPreviousUrl(state.url);
-					this.router.navigate(['/signin']);
-					return false;
+					return this.sessionService.reloadSession();
```
reblace, line 56: I think we should talk through some of these changes.

If the user isn't logged in, the session service is going to trigger the AuthInterceptor to redirect the user to the login page. That makes the first block of `checkAccess` impossible to trigger and redundant (I think). Moreover, a failed call to reloadSession will probably just kill the observable, so checkAccess never gets called.

At that point, should we just try to make session service responsible for ensuring that the user is logged in, so that logic is centralized? I didn't like the fact that I originally had both the AuthGuard and the AuthInterceptor both redirecting to the signin page, but didn't think through it enough to find a better solution.

src/app/core/auth/auth.guard.ts
```diff
@@ -46,19 +51,78 @@ export class AuthGuard implements CanActivate {
...
+			});
+
+			// If there are roles missing then we need to do something
+			if (missingRoles.length > 0) {
+
+				if (!this.authorizationService.hasRole(Role.USER)) {
+					// If the user is missing the user role, they are pending
+					if (state.url !== '/inactive-user') {
+						this.sessionService.setPreviousUrl(state.url);
+						this.router.navigate(['/inactive-user']);
```
reblace: The desire with `access.component.ts` was to create a single component that would take route params indicating what the access problem was. So, we'd always route to the same single endpoint with different route params indicating if the issue was that the user was inactive, unauthorized, etc. You can look at `auth.interceptor.ts` to see an example of what I was thinking.

src/app/core/auth/auth.guard.ts
jrassa: > Moreover, a failed call to reloadSession will probably just kill the observable, so checkAccess never gets called.
Based on my testing, checkAccess still gets called because the failure is being handled.

> If the user isn't logged in, the session service is going to trigger the AuthInterceptor to redirect the user to the login page. That makes the first block of `checkAccess` impossible to trigger and redundant (I think).

I tested without the first block and found that it is still needed.  While the AuthInterceptor will set the redirect, the AuthGuard will still complete before the redirect occurs.  Without the first block, a subsequent check in the AuthGuard will fail and set a different redirect (which will override the one set by AuthInterceptor).  This resulted in the user being redirected to the inactive user page even though they were not signed in yet.

It is also need for scenarios where a user may directly navigate to a page with AuthGuard, but that doesn't make any HTTP calls that would trigger the AuthInterceptor (i.e. /welcome).

> At that point, should we just try to make session service responsible for ensuring that the user is logged in, so that logic is centralized?

I don't think having the reloadSession call in the AuthGuard is the perfect solution and I tried a few other approaches including this, but couldn't get it to work properly.

> I didn't like the fact that I originally had both the AuthGuard and the AuthInterceptor both redirecting to the signin page, but didn't think through it enough to find a better solution.

I don't see any issue with this.  The AuthInterceptor only handles redirects based on errors returned when making HTTP calls.  The AuthGuard still needs to redirect to sign for scenarios like the welcome page which don't make any HTTP calls.

src/app/core/auth/auth.guard.ts
jrassa: This approach works great for the AuthInterceptor where the error details are dynamic and returned from the rest server, although I am not a fan of how all the text appears in the url.  What do you think would be the best way to approach handling more static errors like these?  I don't like the idea of the error text just being inline in the AuthGuard.  I was thinking about storing common static errors in an object in `access.component.ts` that can be used to look up the status/message based on the specified type.

PR #33, opened 2018-11-25
stylelint integration into the build

reblace:
Integrating style (scss) linting into the build.

- Using `stylelint` along with most default options for scss linting
- Made some minor changes to style files to fix some lint issues
- Disabled a couple of the lint rules to accommodate some harder to follow rules
- Added linting as part of the `build` script for npm

One of the bigger changes to styles I made was to remove the `alert` selector from `system-alert.component.scss` because `alert` is not a valid html tag. Since that selector just wrapped the whole scss style, I can't imagine it's going to change anything to have removed it. Especially considering that styles are scoped.


package.json
```diff
@@ -5,12 +5,13 @@
     "ng": "ng",
     "start": "ng serve --proxy-config proxy.conf.json",
     "start:prod": "ng serve --proxy-config proxy.conf.json --prod",
-    "build": "ng build",
-    "build:prod": "ng build --prod",
+    "build": "npm run lint && ng build",
```
brianghig, line 8: @reblace thanks for including these lint tasks with each build. Should we add these details to the top-level README under the NPM Scripts section? https://github.com/asymmetrik/ngx-starter#npm-scripts

brianghig (changes requested):
README docs look like the only thing that's missing for this update. Thanks for incorporating another automated code analysis tool!

package.json
reblace: Yeah, do we want linting to happen with the build? Or, do we only want it as part of a ci build?

reblace: I collapsed both linting operations into the single lint npm script. We'll leave running the lint as part of the build for now.

brianghig (approved):

PR #35, opened 2018-11-30
WIP bootstrap style cleanup

jrassa:
I was reading about [bootstrap theming](https://getbootstrap.com/docs/4.0/getting-started/theming/#variable-defaults) and discovered that some of our bootstrap scss could be simplified by overriding the bootstrap variable defaults vs. redefining styles.  Setting bootstrap styles this way should result in smaller total css.  If we want to proceed with this change, I still need to go through the rest of our bootstrap styles and see what can be replaced with overriding variables.

### DIFF
--- a/src/styles/ngx-bootstrap/_cards.scss
+++ b/src/styles/ngx-bootstrap/_cards.scss
@@ -1,11 +1,6 @@
 @import '../globals';
 @import '../colors';
 
-$card-border-color: map-get($color-background, 'darker');
-$card-border-radius: $border-radius;
-
 .card {
-	border-color: $card-border-color;
-	border-radius: $card-border-radius;
 	box-shadow: 0 1px 1px rgba(0, 0, 0, .3);
 }
--- a/src/styles/_ngx-bootstrap.scss
+++ b/src/styles/_ngx-bootstrap.scss
@@ -1,3 +1,5 @@
+@import './ngx-bootstrap/variables';
+
 @import '../../node_modules/bootstrap/scss/bootstrap';
 
 /**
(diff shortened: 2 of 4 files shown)


src/styles/ngx-bootstrap/_cards.scss
```diff
@@ -1,11 +1,6 @@
 @import '../globals';
 @import '../colors';
 
-$card-border-color: map-get($color-background, 'darker');
-$card-border-radius: $border-radius;
-
 .card {
-	border-color: $card-border-color;
```
brianghig, line 8: @jrassa do you still need to include this use of the SCSS variable in the class definition, or is that handled through some other mechanism?

jrassa: @brianghig 

In the bootstrap scss that we import, there are default styles for everything.  Many of the styles are set using variables that can be overridden.  

For this use case (the border color of cards), you can see it is setting the border color [here](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_card.scss#L13) using a variable named `$card-border-color` which is defined [here](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L796).

Most (all?) of the variables in bootstrap are defined with `!default`.  This means that the value assigned to the variable will be set only if a value has not already been set.  This allows us to simply override the variable vs. redefining the style.  The end result is that when the bootstrap scss is converted into css, the bootstrap styles will now contain our overridden values vs the default values.

jrassa:
Another thing to think about along these lines, is `_display.scss`.  This is where the padding-* and margin-* classes are defined.  Bootstrap provides [similar functionality](https://getbootstrap.com/docs/4.1/utilities/spacing/) backed by a `$spacer` variable.  margin and padding classes are defined as increments of that the spacer to allow for easy global adjustment of spacing.  We may want to consider using these vs. our own homegrown classes.

jrassa:
closing, much of this was handled in #42 and #43

PR #36, opened 2018-12-02
Refactoring paging components to extract common default sorting options

brianghig:
Minor update to extract the default sorting options into the PagingComponent abstract class.

### DIFF
--- a/src/app/common/paging/pager/pager.component.ts
+++ b/src/app/common/paging/pager/pager.component.ts
@@ -2,6 +2,7 @@ import { Component, Input, Output, SimpleChange, EventEmitter, OnInit, OnChanges
 
 import isNumber from 'lodash/isNumber';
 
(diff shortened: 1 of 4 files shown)


jrassa (commented):
@brianghig This looks pretty straightforward to me.  However, if we are going to continue to expand the functionality of `PagingComponent`, we should probably pull it out into its own file.

brianghig:
@jrassa I agree. Before extending functionality in the `PagingComponent` though, I wanted to see if it made sense to pull in any third party libraries that do a better job of managing the paging / sorting experience. Beyond working  Any thoughts from @reblace @mlhummon @ymao1 @mheir or others?

Is this paging component worth iterating on, or do we have something better that we can standardize on? It'd be nice to have something that handles infinite scrolling while still providing a dynamic row or grid configuration.

PR #39, opened 2019-01-26
WIP: Cleanup import paths

abemac:
See https://github.com/Asymmetrik/ngx-starter/issues/38

### DIFF
--- a/src/app/site/example/example-routing.module.ts
+++ b/src/app/site/example/example-routing.module.ts
@@ -1,7 +1,7 @@
 import { NgModule } from '@angular/core';
 import { RouterModule } from '@angular/router';
 
-import { AuthGuard } from '../../core/auth/auth.guard';
+import { AuthGuard } from '@core/auth/auth.guard';
 import { ExploreComponent } from './explore.component';
 import { SearchComponent } from './search.component';
 import { WelcomeComponent } from './welcome.component';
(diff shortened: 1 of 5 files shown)


abemac:
Is everyone okay with the import path aliases I made in tsconfig.json? I want to check before I update all the files

brianghig (approved):

reblace:
I thought we'd previously decided to avoid using aliased paths. My concern is that this means you have to know where to look to find the actual path to the imports. And, you have to recognize the difference between an actual NPM module and an aliased path ('@core' looks like an NPM module namespace). IMO, it hurts the readability of the code.

jrassa:
@reblace Sorry I didn't remember the previous discussion.  @abemac suggested and it seemed like a good idea.  To be clear, this wouldn't change every import statement.  The aliases would only be used when importing across the 3 primary sections (site, core and common).  So, for example, within site, we would still use relative imports for anything else in site, but would use an alias when importing something from core or common.

Would it be better/clearer if we changed the aliases?  Looking around online, I saw several suggestions to use `@app` to indicate it is importing from within the application.   Or we could use `@ngx-starter` to be really explicit.

abemac:
We could create the aliases without the "@". I kind of like having an alias for app, assets, and styles. The imports would look like 
```
import {Something} from 'app/core/something'
```
or 
```
import {Something} from 'app/common/something'
```

This wouldn't conflict with relative imports because relative would start with `./`

PR #41, opened 2019-02-24
Updating NPM dependencies to resolve CVEs

brianghig:


jrassa:
Thanks for taking care of this.

PR #42, opened 2019-03-06
refactoring scss to ease ux implementation

jrassa:
I have been working with UX on updating my app.  This apply some of the lessons learned to the starter to ease future UX implementation.  It relies on the Sass`!default` modifier to make it easier to override variables.

### DIFF
--- a/src/app/common/loading-spinner/loading-spinner.component.scss
+++ b/src/app/common/loading-spinner/loading-spinner.component.scss
@@ -1,10 +1,11 @@
+@import '../../../styles/overrides';
 @import '../../../styles/colors';
 @import '../../../styles/typography';
 
-$loading-spinner-border: 5px;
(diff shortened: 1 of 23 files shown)


reblace:
I have concerns about some of the variable name changes. I tried to establish a convention that made it clear what the variable represented (e.g., ''color-text'), but some of the changes have made it so you won't easily know what the variable actually represents.

As an example, 'color-text' got changed to 'ux-body-text'. The issue with the variable name 'ux-body-text' is that I don't know at a glance what the variable is storing. Is it color? Is it size? etc.

Can you try to modify your naming convention to maintain that sort of hungarian notation naming style? I don't think we need to be super strict about it, but having the names be less vague will help I think.

jrassa:
The variable names have been updated.

jrassa:
I am going to merge this if there are no further concerns.

Issue #16, opened 2018-11-08
Allow for components to register with the Nav Bar


brianghig:
Similar to the Help Topics registration pattern, components like "admin" and each site module should be able to register their navigation items with the nav bar in order to decouple those components from updates to the nav bar directly.

PR #50, opened 2019-05-03
Navbar enhancements

jrassa:
allow components to register with nav bar
enable scrolling of navbar items on small resolutions.
add animation to collapse/expand


jrassa:
Addresses #9 and #16

PR #60, opened 2019-08-22
added focus styles to checkboxes

ktschmincke:


brianghig (approved):
Functionally, this looks good on latest Chrome. Have we tested on older version of Firefox? I'm pushing one update for the Travis config that should resolve the build errors. Thanks for the slick update!

ktschmincke:
@brianghig I can't get the app to load at all in FF 31, even before this change. 
Screen Shot 2019-08-23 at 9 07 54 AM
It's like the typescript didn't transpile or something. Maybe I'm building it wrong? I'm just running `npm start`

brianghig:
@ktschmincke no, I don't think you're building anything wrong. There's a big difference between the dev mode (`npm start`) and the compilation that takes it to backward compatibility (`npm run build:prod` and deploying behind something like Nginx).

Looks like it won't load in dev mode for FF 31 because FF 44 was the first version that supported `let` and many of our more modern tools.

That said, I just built for prod and deployed locally to an Nginx. Confirmed that everything with FF 31.6 looks good, including the slick checkbox animation that you added. Nice work!

ktschmincke:
But shouldn't the typescript compiler translate all the code down to ES5 syntax? e.g. replacing `let` with `var`. We have other tools that use TS (including `let`, `const`, etc) and they run fine on FF 31

ktschmincke:
ah, sorry, I just re read your last comment and understood what you're saying about compiling vs dev mode. I get it now. Thanks for testing for me!

jrassa:
@ktschmincke as part of the upgrade to 8, angular will now generate 2 different bundles, modern (es2015) and legacy (es5).  Only the modern bundle is generated by default when using `ng serve`.

https://angular.io/guide/deployment#local-development-in-older-browsers

PR #62, opened 2019-08-23
added a skip navigation link

ktschmincke:
The skip link appears in the upper left hand corner only when focused and when activated, it skips over the page's navigation links and focuses on the main content of the page.

### DIFF
--- a/src/app/common/directives.module.ts
+++ b/src/app/common/directives.module.ts
@@ -2,16 +2,19 @@ import { NgModule } from '@angular/core';
 
 import { Transclusion } from './directives/transclusion.directive';
 import { NamedTemplate } from './directives/named-template.directive';
+import { SkipToDirective } from './directives/skip-to.directive';
 
 @NgModule({
 	imports: [],
 	exports: [
 		Transclusion,
 		NamedTemplate,
+		SkipToDirective
 	],
 	declarations: [
 		Transclusion,
 		NamedTemplate,
+		SkipToDirective
 	],
 	providers: []
 })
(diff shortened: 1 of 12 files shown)


brianghig (changes requested):
Functionally, this looks good. I'm concerned about using the "id" attribute to identify these elements. It may either overwrite an existing "id" element, or have a collision with another element that also has the skipTo directive applied to it. What do you think about using some other CSS selector mechanism, such as adding a class or using a data-* attribute?

ktschmincke:
Yeah, I had concern about using the id field as well. I originally tried to simply link to the page fragment, e.g. `<a href="#skip-to">Skip to main content</a>`. But this didn't play nice with the router. So I went the way of programmatically changing focus and kept the id.

You're definitely right about potential collisions, especially when multiple components are on the same page and the developer might get confused and add the `skipTo` directive to more than one of them. But using a different selector won't really solve that problem. I used `id` because what I really want is the semantics of the `id` attribute, namely that there can only exist one per page. I'm just not sure about how to enforce that.

Maybe I can build a checker into the directive that will throw an error if it's trying to add the `id` when one already exists. Let me dig into it a bit.

brianghig:
@ktschmincke my main concern is potentially overwriting the 'id' attribute. I think it's acceptable to find however many "skip to" directives are used, and pick the first one. Not sure if it's worth logging a console warning if multiple directives are found, but that could be a way to gracefully handle it.

ktschmincke:
That's fair. Now that I think about it, that would actually make the directive a bit more modular as well. I'll make the change 👍

brianghig (approved):

Issue #64, opened 2019-08-28
Fix Feedback flyout


brianghig:
Upon submitting valid feedback via the flyout option, the app throws a JS error.

brianghig:
Fixed by @tzfx

PR #65, opened 2019-08-28
#64 - Feedback Flyover now allows for optional classification.

tzfx:
Fix for #64 

The feedback service had some message generation based on `feedback.classification.(prefix|level)`, while the feedback modal allowed for optional classification.

- Created a `Classification` class to ensure we set prefix/level.
- Made the `Feedback` class's classification property a `Classification`, and allowed it to be optional.
- Included feedback format tests.

### DIFF
--- a/src/app/core/feedback/feedback.service.spec.ts
+++ b/src/app/core/feedback/feedback.service.spec.ts
@@ -1,5 +1,5 @@
-import { FeedbackService } from './feedback.service.js';
-import { Feedback } from './feedback.model.js';
+import { FeedbackService } from './feedback.service';
+import { Feedback } from './feedback.model';
 
 describe('FeedbackService', () => {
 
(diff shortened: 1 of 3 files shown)


src/app/core/feedback/feedback.service.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { FeedbackService } from './feedback.service.js';
```
brianghig: @tzfx can you remove the trailing `.js` suffix from these imports? Although it works, it's throwing off the code coverage. Removing the suffix (and allowing the import to assume typescript) handles that scenario.

brianghig (changes requested):

src/app/core/feedback/feedback.service.spec.ts
tzfx: C'est terminé.

PR #71, opened 2019-08-30
cleanup search-input-component dom.  add class to improve ability to …

jrassa:
…style from parent components


src/app/common/search-input/search-input.component.ts
```diff
@@ -1,14 +1,13 @@
...
 	templateUrl: './search-input.component.html',
 	styleUrls: ['./search-input.component.scss']
 })
-export class SearchInputComponent implements OnInit {
+export class SearchInputComponent {
 
 	@Input() placeholder: string = 'Search...';
-	@Input() width: string = '500';
 	@Output() applySearch: EventEmitter<string> = new EventEmitter();
 	search: string = '';
```
abemac: Can we also add the ability to set default search? Like making search an Input, or having another input

PR #73, opened 2019-09-05
updates to pageable table component; ux updates to admin tables

jrassa:
Made some updates to PageableTable component.  Cleaned up DOM.  Tweaked inputs/templates.

### DIFF
--- a/src/app/common/paging/pageable-table/pageable-table.component.spec.ts
+++ b/src/app/common/paging/pageable-table/pageable-table.component.spec.ts
@@ -15,11 +15,9 @@ import { PipesModule } from '../../pipes.module';
 	template:
 		`
 	<pageable-table [items]="items"
-		[pagingOptions]="pagingOptions"
-		[refreshable]="refreshable"
-		(onPageChange)="pageChanged$.next($event)"
-		(onRefresh)="refreshed$.next($event)"
-	>
+					[hasItems]="hasItems"
+					[pagingOptions]="pagingOptions"
+					(onPageChange)="pageChanged$.next($event)">
 
 		<ng-template named-template="table-header">
 			{{ headerContent }}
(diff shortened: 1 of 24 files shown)


src/app/common/paging/pageable-table/pageable-table.component.ts
```diff
@@ -15,38 +15,41 @@ import { PagingOptions, PageChange } from '../pager/pager.component';
 })
 export class PageableTable implements AfterContentInit {
 
-	@Input() items: Array<any>;
 	@Input() pagingOptions: PagingOptions = new PagingOptions();
+	@Input() items: Array<any>;
+	@Input() hasItems: boolean = false;
```
tzfx, line 20: I would probably make `hasItems` a method that checks if `this.items` is empty or not instead of letting the parent container decide if items exist out-of-band with the items array.

Doing this also will let us call `hasItems()` instead of using `{{ items?.length === 0 }}` in the template. `items?.length` will actually resolve to `null` in the case that items comes back as null/undefined, which will give us the unintended consequence (`null !== 0`) of unhiding all the stuff that's hiding behind that check.

jrassa: `hasItems` isn't intended to mirror the current state of `items`.  It is supposed to represent the initial state.  It is needed so that we can varying the empty state message based on if there were truly no results or if the user applied an additional filter/search that yielded no results.

src/app/common/paging/pageable-table/pageable-table.component.spec.ts
```diff
@@ -31,24 +29,28 @@ import { PipesModule } from '../../pipes.module';
 			{{ index }}
 		</ng-template>
 
-		<ng-template named-template="empty-table">
-			{{ emptyTableContent }}
+		<ng-template named-template="table-no-data">
```
brianghig, line 32: @jrassa is this intended to represent an error state? Otherwise, what's the distinction between "table-no-data" and "table-no-results" ?

brianghig: Nevermind.. I saw your note below. Probably worth documenting either inline via comments or in the README that we can/should expand :-)

brianghig (approved):

PR #74, opened 2019-09-09
13 Audit Log Module

brianghig:
Beginnings of the Audit Log module. Providing several display types and base test coverage of the audit listing component.

* [x] Move Audit to its own non-admin page
* [x] Move sortObjectKeys to pipe


src/app/core/admin/audit/audit-view-change.component.ts
```diff
@@ -0,0 +1,52 @@
...
+export class AuditViewChangeModal extends AuditViewDetailModal {
+
+	constructor(
+		public modalRef: BsModalRef
+	) {
+		super(modalRef);
+	}
+
+	// Derived from http://stackoverflow.com/a/1359808 and http://stackoverflow.com/a/23124958
+	sortObjectKeys(obj: any): any {
```
tzfx: This looks like a good candidate to in a util class somewhere that's injectable and sharable across the app.

src/app/core/admin/audit/audit.component.ts
```diff
@@ -0,0 +1,271 @@
...
+import { PagingOptions, SortDisplayOption, SortDirection, SortableTableHeader, TableSortOptions, PagingResults } from '../../../common/paging.module';
+import { AdminUsersService } from '../user-management/admin-users.service';
+import { AuditOption } from './audit.classes';
+import { AdminTopics } from '../admin-topic.model';
+
+
+@Component({
+	selector: 'audit',
+	styleUrls: ['audit.scss'],
+	templateUrl: './audit-list.component.html'
```
tzfx: Probably want this to be `audit.component.html` to clearly link this component and template.

jrassa: I moved and renamed all corresponding files to`list-audit-entries`

jrassa:
Should this be under admin?  In the old starter it was its own separate page that was only accessible to users with the 'Auditor' role.  The backend is still enforcing this restriction.  When first testing I got an error since I did not have the required role.

src/app/core/admin/audit/audit-view-change.component.ts
brianghig: @tzfx Agreed! I think it makes sense to be its own pipe, since it's applicable to the templates. Will change.

brianghig:
> Should this be under admin? In the old starter it was its own separate page that was only accessible to users with the 'Auditor' role. The backend is still enforcing this restriction. When first testing I got an error since I did not have the required role.

@jrassa no, it should probably be under its own "audit" section. I'll make that change now.

brianghig:
@tzfx @jrassa I believe that all comments thus far have been addressed. Thanks for the feedback!

Issue #30, opened 2018-11-15
Migrate from pull-left and pull-right


brianghig:
To Bootstrap’s new float-left and float-right

jrassa:
resolved by #31

PR #75, opened 2019-09-10
BUG: set initial admin when creating team

abemac:
Currently, initial admin will be current user, regardless of who is specified.

### DIFF
--- a/src/app/core/teams/teams.service.ts
+++ b/src/app/core/teams/teams.service.ts
@@ -42,7 +42,10 @@ export class TeamsService {
 	create(team: Team, firstAdmin?: any): Observable<any> {
 		return this.http.put(
 			`api/team`,
-			JSON.stringify(team),
+			JSON.stringify({
+				team: team,
+				firstAdmin: null === firstAdmin ? null : firstAdmin._id || firstAdmin
+			}),
 			{ headers: this.headers }
 		).pipe(
 			catchError((error: HttpErrorResponse) => {
--- a/src/app/core/teams/teams.service.ts
+++ b/src/app/core/teams/teams.service.ts
@@ -39,12 +39,12 @@ export class TeamsService {
 	// 	return this.asyHttp.put(new HttpOptions('team', () => { }, { team, firstAdmin }));
 	// }
 
-	create(team: Team, firstAdmin?: any): Observable<any> {
+	create(team: Team, firstAdmin?: string): Observable<any> {
 		return this.http.put(
 			`api/team`,
 			JSON.stringify({
 				team: team,
-				firstAdmin: null === firstAdmin ? null : firstAdmin._id || firstAdmin
+				firstAdmin: firstAdmin ? firstAdmin : null
 			}),
 			{ headers: this.headers }
 		).pipe(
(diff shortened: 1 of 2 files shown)


jrassa:
Are you working on a corresponding PR for node-rest-starter?  As is an error is now thrown when trying to create a team.

abemac:
@jrassa yes, I am. https://github.com/Asymmetrik/node-rest-starter/pull/30/

src/app/core/teams/teams.service.ts
```diff
@@ -42,7 +42,10 @@ export class TeamsService {
 	create(team: Team, firstAdmin?: any): Observable<any> {
 		return this.http.put(
 			`api/team`,
-			JSON.stringify(team),
+			JSON.stringify({
+				team: team,
+				firstAdmin: null === firstAdmin ? null : firstAdmin._id || firstAdmin
```
brianghig: @abemac can we please get a test for this logic? Looking at it, I'm not sure whether the `||` applies to the full statement or just the last two parameters. Might be nice to type-check "firstAdmin" as a `string | User` too. Thanks for fixing this bug!

brianghig (changes requested):

src/app/core/teams/teams.service.ts
jrassa: Can we simplify things and just make "firstAdmin" a `string` and ensure that the id string is always passed in?

brianghig: @jrassa @abemac  That would certainly work. Does it make sense to link this by User ID or by Username? My concern with linking by User ID is that a user's profile can be destroyed and recreated (mostly in the PKI scenario), and all of their Team Admin Ownership would be lost.

jrassa: I'm not really sure what scenario would cause a user's profile to be destroyed.  When using PKI, if a user's PKI dn were to change a new profile would be created.  In that scenario, I would expect that the user would contact support and the accounts could be merged/reconciled.

brianghig: In my experience, we don't do a good job of managing new DNs with the same Usernames. It causes conflicts when we pull cache entries by DN, and try to sync that up with DB records by username.

In general, I'd rephrase this to a question of, "how should we uniquely identify users?" Username seems like a common platform between "local" and "pki" strategies, especially since it is how users and administrators search for User-based content.

jrassa: >In my experience, we don't do a good job of managing new DNs with the same Usernames. It causes conflicts when we pull cache entries by DN, and try to sync that up with DB records by username.

I believe that we have improved on this recently.  https://github.com/Asymmetrik/node-rest-starter/commit/d9091757b3106aa380ac938cc9f33660b5577e3d#diff-1687462e99fc8498e54065d21f1f5599L128

>In general, I'd rephrase this to a question of, "how should we uniquely identify users?" Username seems like a common platform between "local" and "pki" strategies, especially since it is how users and administrators search for User-based content.

That is a detail of the user management strategy that the rest of the application shouldn't need to be aware.  Using generated ids, as we are now, decouples the rest of the application from the details of user management.

PR #77, opened 2019-09-16
dom/style updates for help/breadcrumbs

jrassa:


src/app/common/breadcrumb/breadcrumb.component.html
```diff
@@ -1,10 +1,9 @@
...
-	<div *ngSwitchCase="false" class="header">{{ _homeBreadcrumb?.label }}</div>
-	<div *ngSwitchCase="true" class="header">
-		<span *ngFor="let b of breadcrumbs; let last = last" class="header-breadcrumb">
+<div class="header">
+	<ng-container *ngFor="let b of breadcrumbs; let last = last" class="header-breadcrumb">
+		<span class="header-breadcrumb">
 			<a *ngIf="!last" [routerLink]="[b.url]">{{ b?.label }}</a>
-			<span *ngIf="last">{{ b?.label }}</span>
-			<span *ngIf="!last" class="header-breadcrumb-separator"> > </span>
+			<ng-container *ngIf="last">{{ b?.label }}</ng-container>
```
tzfx, line 5: I'd suggest doing something like `<span class="last" *ngIf="last">{{ b?.label }}</span>` instead of the ng-container so the user can style the final/active breadcrumb link if they'd like via .header-breadcrumb.last or whatever.
Otherwise, this all looks good.

jrassa: The last breadcrumb is already in a span.  Below is a cleaned up sample of the generated html.

```html
<div class="header">
	<span class="header-breadcrumb"><a ng-reflect-router-link="/help" href="#/help">Help</a></span>
	<span class="header-breadcrumb-separator"> &gt; </span>
	<span class="header-breadcrumb">Getting Started</span>
</div>
```

Styling the last breadcrumb can be done using a pseudo-class.
```css
.header-breadcrumb:last-child {
	color: red;
}
```

tzfx: Oh jeez- Yeah think I saw the unified dif and lost track of the spans. This looks good.

tzfx (approved):

PR #78, opened 2019-09-16
add getTeam method to return all teams a user can see

jrassa:


src/app/core/teams/teams.service.ts
```diff
@@ -137,27 +137,24 @@ export class TeamsService {
...
-				(err: any) => {
-					observer.error(err);
-				},
-				() => {
-					observer.complete();
-				}
-			);
-		});
+		return this.getTeams().pipe(
+			map((teams: Team[]) => {
```
tzfx, line 151: :+1: Much cleaner.

tzfx:
LGTM 👍 💯

PR #81, opened 2019-09-19
use clientUrl vs baseUrl; fix setting of currentRoute for old feedback modal

jrassa:
Small changed due to Asymmetrik/node-rest-starter#34

Also noticed that fix for feedback currentRoute was applied only to feedback flyout, not the modal.


tzfx:
Good catch :+1:

PR #82, opened 2019-09-20
tslint update

jrassa:
Looking at leveraging tslint more.  I updated the rules to more closely match those generated by ng cli when creating a new project (which match up with the Angular style guide).  This includes additional Angular specific rules from codelyzer.  Rule changes are up for discussion if anyone feels strongly for/against any of the changes.

For now, I changed any rules that were generating errors to generate warnings.  That way downstream projects won't be forced to fix a bunch of lint errors immediately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@types/jasmine": "2.8",
     "@types/jasminewd2": "2.0",
     "@types/node": "8.9",
-    "codelyzer": "^5.0.1",
+    "codelyzer": "^5.1.1",
     "jasmine-core": "2.99",
     "jasmine-spec-reporter": "4.2",
     "karma": "^4.1.0",
--- a/src/app/common/breadcrumb/breadcrumb.service.ts
+++ b/src/app/common/breadcrumb/breadcrumb.service.ts
@@ -8,7 +8,7 @@ export interface Breadcrumb {
 export class BreadcrumbService {
 
 	static getBreadcrumbs(route: ActivatedRouteSnapshot, url: string = '', breadcrumbs: Breadcrumb[] = []): Breadcrumb[] {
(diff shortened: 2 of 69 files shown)


brianghig:
@jrassa thanks for bringing us closer to tslint standards! IMO, we should leave most / all of these as errors unless they're causing functional issues. Having clean, consistent coding guidelines seems to help the transition between projects. Otherwise, are we left with a bunch of warning logs that folks would tend to ignore over time?

jrassa:
@brianghig I agree that we should set them as errors.  I was mainly considering having a short transition period of maybe a couple of weeks at most before we switch to errors.  That would allow us time to make changes to our upstream projects without being forced to fix everything immediately.  I am happy to enable them now though.

brianghig:
@jrassa whew! Lots of changes! Thanks for tackling this, especially with the "rip the bandaid" approach.

jrassa:
I fixed most of the warnings and switched them to be errors.  Many can be autofixed using `ng lint --fix`.  I left a few as warnings as fixing them will be a bit more invasive (renaming things).  I will prepare a follow up PR to handle those changes.

brianghig:
Makes sense. There's one error that's preventing me from building / starting the app at the moment:

```
ERROR in src/app/core/admin/end-user-agreement/admin-edit-eua.component.ts(13,14): error TS2420: Class 'AdminUpdateEuaComponent' incorrectly implements interface 'OnDestroy'.
  Property 'ngOnDestroy' is missing in type 'AdminUpdateEuaComponent' but required in type 'OnDestroy'.
```

brianghig:
Pending a TravisCI successful completion, I'm good with this update. Ran it locally to verify, and all seemed fine.

brianghig (approved):

PR #85, opened 2019-09-23
added non-html5 browser support for sectioning elements

ktschmincke:
per https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines#Using_HTML5_elements_in_non-HTML5_browsers

this will allow us to use html5 sectioning elements (`<header>`, `<nav>`, `<article>`, etc) even in browsers that do not support them

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -29,10 +29,6 @@ html {
 	min-height: 100%;
 }
 
-article, aside, footer, header, nav, section {
-	display: block;
-}
-
 @if $form-required == before {
 	.form-required:before {
 		content: '*';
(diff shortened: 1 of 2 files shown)


src/styles.scss
```diff
@@ -29,6 +29,10 @@ html {
 	min-height: 100%;
 }
 
+article, aside, footer, header, nav, section {
```
jrassa: Not sure this is needed.  bootstrap styles that are imported above includes the follow:

```css
article, aside, figcaption, figure, footer, header, hgroup, main, nav, section {
  display: block;
}
```

ktschmincke: good catch. i'll remove

PR #90, opened 2019-10-04
Flyout position support

jrassa:
Updates base flyout component with support for setting position (right, left, top, bottom).
Minor updates to feedback flyout due to changes to flyout component.


tzfx (approved):
👍 
> I'm commander Shepard and this is my favorite merge request on the citadel.

PR #91, opened 2019-10-07
Add support for expandable/collapsable rows in pageable-table

jrassa:


brianghig:
Code inspection looks good. The build check passed, but it didn't reflect in the PR.

PR #117, opened 2020-03-19
feat: add option to display table actions even if table has no items

sshah-asymmetrik:
Hi John!

This PR is to add an option to the pageable-table.component that allows us to display table actions even if there are no items.

I tried to write it and set defaults so that any current users of the pageable-table won't see any differences.

Let me know if there are any additional changes I should make to get this merged.

Thanks!

### DIFF
--- a/src/app/common/paging/pageable-table/pageable-table.component.html
+++ b/src/app/common/paging/pageable-table/pageable-table.component.html
@@ -1,23 +1,23 @@
 <div class="table" [ngClass]="{'card p-3': showInCard}">
-	<div class="table-header align-items-center mb-3" *ngIf="pagerAtTop || showActions" [hidden]="itemsInternal?.length === 0">
-		<div class="table-actions d-flex flex-grow-1" *ngIf="showActions">
+	<div class="table-header align-items-center mb-3" *ngIf="pagerAtTop || showActions">
+		<div class="table-actions d-flex flex-grow-1" *ngIf="showActions" [hidden]="itemsInternal?.length === 0 && hideActionsNoData">
 			<ng-container *ngTemplateOutlet="actionTemplate"></ng-container>
 		</div>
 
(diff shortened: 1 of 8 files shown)


src/app/common/paging/pageable-table/pageable-table.component.html
```diff
@@ -61,13 +61,13 @@
 
 		<div class="table-footer-pager" *ngIf="pagerAtBottom" [ngClass]="{'ml-auto': !showFooterActions}">
```
jrassa: Can we apply this to the footer actions as well?

src/app/common/paging/pageable-table/pageable-table.component.ts
```diff
@@ -14,8 +14,9 @@ export class PageableTableComponent {
 	@Input() loading = false;
 	@Input() showInCard = false;
 	@Input() showActions = false;
+	@Input() hideActionsNoData = true;
```
jrassa: I think we should set the default to `false`.  That will make it's use consistent with `hideTableNoData`.  I think our 2 downstream projects are the only ones we have to worry about for compatibility and I am not actually using table actions anywhere currently.

jrassa (changes requested):
This is a really good idea.  I actually avoided using the table actions in a few places because this option didn't exist.

jrassa:
Also, if you wanted to, the admin pages could be updated to take advantage of this.

src/app/common/paging/pageable-table/pageable-table.component.html
sshah-asymmetrik: Good call! I added it as a separate option for footer actions, `hideFooterActionsNoData`.

src/app/common/paging/pageable-table/pageable-table.component.ts
sshah-asymmetrik: Great! It is done.

sshah-asymmetrik (commented):
I went ahead and shifted the actions in the admin view tables into tableActions.

PR #119, opened 2020-03-25
build: Update to Angular 9

jrassa:


jrassa:
After merging this into downstream projects, migrations can be rerun to update downstream specific code.

```
ng update @angular/cli --from 8 --to 9 --migrate-only
ng update @angular/core --from 8 --to 9 --migrate-only
```

jrassa:
Updated to 9.1.

PR #120, opened 2020-03-27
Feat/user preferences template

sshah-asymmetrik:
Hey John,

This branch is for adding the ability to inject custom component templates into the site container. The context for this is that we're trying to add a User Preferences page for chainexplorer. I figure that every application is likely to have unique preferences that users may want to set, so I was trying to come up with a way for integrating custom pieces. 

With the solution presented on this branch currently, to supply a template to the site container the application would have to modify it's own `app.component.html` to look something like:
```
<div class="app-container">

	<site-container [userPreferencesTemplate]="userPreferencesTemplate">

		<router-outlet></router-outlet>

	</site-container>

</div>

<ng-template #userPreferencesTemplate>
	<ce-user-preferences></ce-user-preferences>
</ng-template>
``` 
and also the `app.module.ts` to import the site's so that we can use that templated component.

There's a lot of little additions in the user directory, which is because I was trying to replicate the structure in the other directories with the routing and modules and whatnot, especially if over time there's more adding of user-related stuff to the starter.

This was kind of a shot in the dark for me, but it seems to work. Let me know if you have any ideas on how to improve this!

Thanks!


jrassa:
Generally, we try to come up with a way to allow downstream projects to avoid having to modify any files in the starter.  I think we can do that here.

There are 2 possible scenarios to consider.  The first and more complicated one would be if we want to combine some generic user preferences defined in the starter with app specific preferences defined in the downstream application.  Your solution seems to be leaning in this direction, but currently we don't have any generic preferences to define.  However, if we wanted to explore this, I would suggest a solution similar to how we handle admin pages.

The second and simplier one would be to only have downstream app specific preferences.  There are 2 approaches we could take to achieve this.

1. Add a config option that specifies the router path (ex: user/userPreference).  If path is present in config, then render preference option in navbar.  Downstream app is responsible for definition component/route.
2. Come up with a way to programmatically add items to user nav popover menu.  Possibly by extending existing NavbarTopic class.

I would probably recommend going with option 1.  It is the simpliest option and I can't currently think of any additional use cases where we would need the flexibility to add additional items to the user nav popover.

jrassa:
Thanks.

PR #132, opened 2020-05-20
feat: Initialize MessageService for valid users only

jrassa:


src/app/core/messages/message.service.ts
```diff
@@ -20,11 +23,17 @@ export class MessageService {
...
 
 	constructor(
+		private sessionService: SessionService,
+		private authorizationService: AuthorizationService,
 		private alertService: SystemAlertService,
 		private http: HttpClient,
 		private socketService: SocketService
 	) {
-		this.initialize();
+		this.sessionService.getSession().subscribe((session: Session) => {
```
tzfx: This seems-- odd. Does the session service.getSession drive the state of authorizationService?
It looks like the authorizationService itself already should have the session in its own constructor-- Can we not just check `authorizationService.isUser()` directly without doing anything with the session service?

jrassa: There is no guarantee that the session will have been loaded and available for the authorizationService when the message service is created.  Subscribing to the session ensures that we defer the check until the we know for sure that the session is available.

Also helpful when using basic auth.  Initial session will be null.  Once user signs in, session is updated.

PR #133, opened 2020-05-27
css: square button support.

tzfx:
This change adds the `btn-square` css class to:

- `btn`
- `btn-sm`
- `btn-lg`

This class uses uniform button padding all-the-way-around on the default/sm/large button elements, which allows for more accurate centering of things like font-awesome icons in conjunction with the `fa-fw` fixed-width class.


tzfx:
Erk. I think I missed a file.

jrassa (approved):
Looks good.  I added examples to the form component.  Also fixed a bug where it was using `btn-large` vs. `btn-lg`.

PR #140, opened 2020-07-13
fix: Ensure Observable subscriptions are unsubscribed to prevent memory leaks

jrassa:
Memory leaks can be caused by Observable subscriptions that are not properly unsubscribed.  Using `UntilDestroy` to ensure subscriptions are unsubscribed when containing class is destroyed.

### DIFF
--- a/src/app/core/about.component.ts
+++ b/src/app/core/about.component.ts
@@ -1,8 +1,11 @@
 import { Component, OnInit } from '@angular/core';
 
+import { untilDestroyed, UntilDestroy } from '@ngneat/until-destroy';
+import { first } from 'rxjs/operators';
 import { Config } from './config.model';
 import { ConfigService } from './config.service';
 
+@UntilDestroy()
 @Component({
 	template: `
 		<div class="container">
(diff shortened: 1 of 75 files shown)


src/app/core/about.component.ts
```diff
@@ -23,9 +26,12 @@ export class AboutComponent implements OnInit {
 	constructor(private configService: ConfigService) {}
 
 	ngOnInit() {
-		this.configService.getConfig().subscribe((config: Config) => {
-			this.appTitle = config?.app?.title ?? '';
-			this.version = config.version;
-		});
+		this.configService
+			.getConfig()
+			.pipe(first(), untilDestroyed(this))
```
tzfx, line 31: I believe that `first()` (and `take(1)`) should generally unsubscribe once the first observable has been emitted, so untilDestroyed may be redundant here.

tzfx:
This is a pretty cool feature that lets us be just a little bit DRYer! Nice find.

tzfx (approved):

PR #144, opened 2020-07-28
Sonarqube code smells

sshah-asymmetrik:
Fixing all of the SonarQube bugs/code smells detected. They're all pretty minor things. Cleaning up commented code, adding bits to html tags to help screen readers, removing unused imports, etc.

The one thing left that SonarQube is mad about is a RegEx in string-utils.service.ts (line 69) that it views as a security hotspot.

### DIFF
--- a/src/app/common/paging/quick-column-toggle/quick-column-toggle.component.ts
+++ b/src/app/common/paging/quick-column-toggle/quick-column-toggle.component.ts
@@ -1,6 +1,6 @@
 import { ChangeDetectionStrategy, Component } from '@angular/core';
-import { UntilDestroy } from '@ngneat/until-destroy';
 
+import { UntilDestroy } from '@ngneat/until-destroy';
 import { QuickFiltersComponent } from '../quick-filters/quick-filters.component';
 
 @UntilDestroy()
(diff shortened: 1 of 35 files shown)


src/app/common/paging/quick-column-toggle/quick-column-toggle.component.ts
```diff
@@ -29,24 +29,25 @@ export class QuickColumnToggleComponent extends QuickFiltersComponent {
 	private checkColumnConfiguration() {
 		// Check first to see if all columns are turned on
 		this.columnMode = 'all';
-		this.filterKeys.some((key: string) => {
```
jrassa, line 32: What was the issue that sonarqube flagged here?  The update changes the logical flow.  The `return true;` below was returning out of the inline function executed by the iterator.   With this change it is now returning out of the `checkColumnConfiguration` function.  This has the negative side effect that it will skip the logic below that checks if the default columns are enabled.

jrassa: Was it due to the `some` call having side effects (setting `columnMode`)?  The logic flow issue could be fixed by changing the return to a break, but here is a better more concise solution.

```ts
private checkColumnConfiguration() {
	if (this.filterKeys.every(key => this.filters[key].show)) {
		this.columnMode = 'all';
	} else if (
		this.filterKeys.some(key => this.filters[key].show !== this.defaultFilters[key].show)
	) {
		this.columnMode = 'custom';
	} else {
		this.columnMode = 'default';
	}
}
```

sshah-asymmetrik: This got flagged as a minor issue because the return value of the `some` call was being ignored. Good catch on the logical flow change! I've replaced the function with the new one you provided.

Issue #151, opened 2020-09-01
Message title should reflect actual value vs. capital case


brianghig:
When a message is added to the system (whether it's a MOTD, INFO, WARN, etc.), the value sent to and returned from the service reflects a certain capitalization scheme entered by the user.

However, when this value is translated into the title of the notification (both in the sidebar flyout and the messages listing grid), the title is converted into capital case.

For example, a title of "TEST Testing" is displayed to the user as "Test Testing".

**Expected Outcome:**
- Notification titles display with the capitalization that the admin entered.

brianghig:
@jrassa @tzfx do you concur with this issue? Any objections?

PR #153, opened 2020-09-02
#151 Removing titlecase from recent and all messages listing

brianghig:
Also adding spec coverage for both components


tzfx (approved):
Nice tests! 👍

Issue #152, opened 2020-09-02
Enable Export from Admin Users listing


brianghig:
Current export is a no-op. We should enable the ability to export the current view, as well as a single field.

PR #154, opened 2020-09-03
#152 Feature: Export Users

brianghig:
This change removes the no-op dropdown for "Export Current View" and "Export Single Field" from the users listing in favor of a single Export button that derives its configuration from the current table layout. The `node-rest-starter` already supports this functionality.

### DIFF
--- a/src/app/core/admin/user-management/admin-list-users.component.ts
+++ b/src/app/core/admin/user-management/admin-list-users.component.ts
@@ -18,8 +18,8 @@ import { first } from 'rxjs/operators';
 import { Role } from '../../auth/role.model';
 import { User } from '../../auth/user.model';
 import { ConfigService } from '../../config.service';
-import { AdminUsersService } from './admin-users.service';
 import { ExportConfigService } from '../../export-config.service';
+import { AdminUsersService } from './admin-users.service';
 
 @UntilDestroy()
 @Component({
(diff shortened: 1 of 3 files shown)


src/app/core/admin/user-management/admin-list-users.component.ts
```diff
@@ -159,11 +161,38 @@ export class AdminListUsersComponent extends AbstractPageableDataComponent<User>
...
+		this.exportConfigService
+			.postExportConfig('user', {
+				q: this.getQuery(),
+				s: this.search,
+				cols: viewColumns,
+				sort: this.pagingOptions.sortField,
+				dir: this.pagingOptions.sortDir
+			})
+			.subscribe((response: any) => {
+				window.open(`/api/admin/users/csv/${response._id}`);
```
tzfx, line 190: Interesting-- so it looks does essentially a post-redirect-get and defers the CSV generation to the server? If the client already has a given display window with data, is there a reason that we aren't generating the CSV in the client code?

I could definitely see us moving toward being able to export every pageable table to a downloaded file, and I'm trying to think of the maintenance lift to have to add a new export endpoint for every table/list object.

brianghig: I think the goal is to be able to export all pages of data. In that case, we need to go to the server to get all of that information.

tzfx: Seems reasonable.

brianghig: Thanks @tzfx ! I just added some spec coverage for this component and the export functionality.

jrassa (approved):
Looks good.  Made one small tweek to build the role columns dynamically to automatically handle when applications add/change roles.

PR #155, opened 2020-09-04
feat: add loading overlay component

sumeetsshah:
This PR is to add a loading-overlay component to the starter. This component is meant to be a semi-opaque overlay for tables/graphs/areas of the page that displays the loading spinner while they're loading. In the event of a loading failure, the component will display an error and allow the users to retry.

The new components are the `loading-overlay` component itself as well as the `notification` component, which is used to display the message.

Also, what should go in the unit tests for the `loading-overlay` component? the component is just a wrapper that combines the `notification` and `loading-spinner` components, each of which has its own unit tests. Should I just test that the `*ngIf`s are working properly?

Thanks!


sumeetsshah:
Using wrong github account. switching to asym account.

PR #156, opened 2020-09-04
feat: add loading overlay component

sshah-asymmetrik:
This PR is to add a loading-overlay component to the starter. This component is meant to be a semi-opaque overlay for tables/graphs/areas of the page that displays the loading spinner while they're loading. In the event of a loading failure, the component will display an error and allow the users to retry.

The new components are the loading-overlay component itself as well as the notification component, which is used to display the message.

Also, what should go in the unit tests for the loading-overlay component? the component is just a wrapper that combines the notification and loading-spinner components, each of which has its own unit tests. Should I just test that the *ngIfs are working properly?

Thanks!

### DIFF
--- a/src/app/common/loading-overlay.module.ts
+++ b/src/app/common/loading-overlay.module.ts
@@ -1,0 +1,13 @@
+import { NgModule } from '@angular/core';
+
+import { LoadingOverlayComponent } from './loading-overlay/loading-overlay.component';
+import { LoadingSpinnerModule } from './loading-spinner.module';
+import { NotificationModule } from './notification.module';
+
+@NgModule({
+	imports: [NotificationModule, LoadingSpinnerModule],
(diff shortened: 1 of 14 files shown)


jrassa:
I'm trying to test this and it doesn't seem to be working.  Could you add an example usage under `src/app/site/example`

> Also, what should go in the unit tests for the loading-overlay component? the component is just a wrapper that combines the notification and loading-spinner components, each of which has its own unit tests. Should I just test that the *ngIfs are working properly?

Yes, test *ngIfs.

sshah-asymmetrik:
@jrassa Thanks for the feedback! I have added an example page for the loading overlay (can be navigated to in the starter from the last icon in the navbar) as well as some unit tests for the overlay. Let me know what you think!

jrassa (approved):
Looks good.

PR #164, opened 2020-10-09
change interceptors to be chainable with bypass

sshah-asymmetrik:
The changes in this PR are to enable a more customizable auth interceptor chain. The idea is to be able to chain together interceptors on top of the default interceptor and, if necessary, bypass interceptors lower down on the chain and pass the error though.

I separated out the signin and eua interceptor code into individual interceptors to serve as an illustration of how we can chain these together and be examples for what other custom interceptors that people add to their application modules could look like.

### DIFF
--- a/src/app/core/auth/abstract.interceptor.ts
+++ b/src/app/core/auth/abstract.interceptor.ts
@@ -15,9 +15,7 @@ export abstract class AbstractHttpInterceptor implements HttpInterceptor {
 	intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
 		return next.handle(req).pipe(
 			catchError(err => {
-				if (!err.bypassAuthInterceptors) {
-					this.handleError(err, req);
-				}
+				this.handleError(err, req);
 				return throwError(err);
 			})
 		);
--- a/src/app/core/auth/abstract.interceptor.ts
+++ b/src/app/core/auth/abstract.interceptor.ts
@@ -6,17 +6,17 @@ import { throwError, Observable } from 'rxjs';
 import { catchError } from 'rxjs/operators';
 
 /**
- * HTTP Interceptor
+ * Abstract HTTP Interceptor
  */
 @Injectable()
-export abstract class Interceptor implements HttpInterceptor {
+export abstract class AbstractHttpInterceptor implements HttpInterceptor {
 	protected constructor(public router: Router) {}
 
 	intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
 		return next.handle(req).pipe(
 			catchError(err => {
 				if (!err.bypassAuthInterceptors) {
-					this.handleError(err);
+					this.handleError(err, req);
 				}
 				return throwError(err);
 			})
@@ -36,14 +36,7 @@ export abstract class Interceptor implements HttpInterceptor {
 	}
 
 	/**
-	 * Set a flag in the error to alert future interceptors to pass the error through
-	 */
-	bypassRemainingInterceptors(err: any) {
-		err.bypassAuthInterceptors = true;
-	}
-
-	/**
 	 * Implement this in any interceptors extending the AuthInterceptor class
 	 */
-	abstract handleError(err: any): void;
+	abstract handleError(err: any, req: HttpRequest<any>): void;
 }
--- a/src/app/core/auth/auth.interceptor.ts
+++ b/src/app/core/auth/auth.interceptor.ts
@@ -1,17 +1,19 @@
+import { HttpRequest } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 
-import { Interceptor } from './interceptor';
+import { AbstractHttpInterceptor } from './abstract.interceptor';
 
 /**
  * HTTP Interceptor that will interpret authentication related HTTP calls
  */
 @Injectable()
-export class AuthInterceptor extends Interceptor {
-	handleError(err: any): void {
-		const { status, type, message, url } = this.parseError(err);
-		if (status === 403) {
-			this.router.navigate(['/access'], { state: { status, type, message, url } });
-			this.bypassRemainingInterceptors(err);
+export class AuthInterceptor extends AbstractHttpInterceptor {
+	handleError(err: any, req: HttpRequest<any>): void {
+		if (!req.headers.has('bypass-auth-interceptor')) {
+			const { status, type, message, url } = this.parseError(err);
+			if (status === 403) {
+				this.router.navigate(['/access'], { state: { status, type, message, url } });
+			}
 		}
 	}
 }
--- a/src/app/core/auth/auth.interceptor.ts
+++ b/src/app/core/auth/auth.interceptor.ts
@@ -1,51 +1,17 @@
-import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest } from '@angular/common/http';
 import { Injectable } from '@angular/core';
-import { Router } from '@angular/router';
 
-import { throwError, Observable } from 'rxjs';
-import { catchError } from 'rxjs/operators';
+import { Interceptor } from './interceptor';
 
 /**
- * HTTP Interceptor that will interpret authentication-related HTTP calls
+ * HTTP Interceptor that will interpret authentication related HTTP calls
  */
 @Injectable()
-export class AuthInterceptor implements HttpInterceptor {
-	constructor(public router: Router) {}
-
-	intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
-		return next.handle(req).pipe(
-			catchError(err => {
-				if (!err.bypassAuthInterceptors) {
-					this.handleError(err);
-				}
-				return throwError(err);
-			})
-		);
-	}
-
-	/**
-	 * Grab all the useful stuff out of the error response
-	 */
-	parseError(err: any) {
-		const status = err?.status ?? 200;
-		const type = err?.error?.type ?? '';
-		const url = err?.url ?? '';
-		const message = err?.error?.message ?? '';
-
-		return { status, type, message, url };
-	}
-
-	/**
-	 * Set a flag in the error to alert future interceptors to pass the error through
-	 */
-	bypassRemainingInterceptors(err: any) {
-		err.bypassAuthInterceptors = true;
-	}
-
-	/**
-	 * Default error handler. Override this in any interceptors extending the AuthInterceptor class
-	 */
+export class AuthInterceptor extends Interceptor {
 	handleError(err: any): void {
-		this.router.navigate(['/access'], { state: this.parseError(err) });
+		const { status, type, message, url } = this.parseError(err);
+		if (status === 403) {
+			this.router.navigate(['/access'], { state: { status, type, message, url } });
+			this.bypassRemainingInterceptors(err);
+		}
 	}
 }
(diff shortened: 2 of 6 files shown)


src/app/core/auth/auth.interceptor.ts
```diff
@@ -4,41 +4,48 @@ import { Router } from '@angular/router';
 
 import { throwError, Observable } from 'rxjs';
 import { catchError } from 'rxjs/operators';
-import { SessionService } from './session.service';
```
sshah-asymmetrik, line 7: Removed unused import

src/app/core/auth/auth.interceptor.ts
```diff
@@ -4,41 +4,48 @@ import { Router } from '@angular/router';
...
 export class AuthInterceptor implements HttpInterceptor {
-	constructor(private router: Router, private sessionService: SessionService) {
-		// Nothing here
-	}
+	constructor(public router: Router) {}
 
 	intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
 		return next.handle(req).pipe(
 			catchError(err => {
-				// Grab all the useful stuff out of the error response
```
sshah-asymmetrik, line 21: Wrapped the error parsing in the `parseError` method

src/app/core/auth/eua.interceptor.ts
```diff
@@ -0,0 +1,11 @@
+import { AuthInterceptor } from './auth.interceptor';
+
+export class EuaAuthInterceptor extends AuthInterceptor {
```
brianghig: @sshah-asymmetrik do these subclasses need the `@Injectable` annotation, or does the parent class cover that?

src/app/core/core.module.ts
```diff
@@ -92,7 +94,11 @@ export function getConfiguration(configService: ConfigService) {
 			deps: [ConfigService],
 			multi: true
 		},
-		{ provide: HTTP_INTERCEPTORS, useClass: AuthInterceptor, multi: true }
+		[
+			{ provide: HTTP_INTERCEPTORS, useClass: AuthInterceptor, multi: true },
```
brianghig, line 98: @sshah-asymmetrik what guarantees the order of interceptors if we're decorating the error with an attribute that needs to be respected by subsequent interceptors?

src/app/core/auth/eua.interceptor.ts
sshah-asymmetrik: @brianghig They seem to work without the annotation, which makes me think they're inheriting it from the parent class. That being said, I think it's better to include the annotation anyways. It doesn't do any harm and can serve as preparation for if the interceptor were changed to be a standalone that doesn't extend anything.

src/app/core/core.module.ts
sshah-asymmetrik: @brianghig  Angular itself should be guaranteeing that order. Interceptors are executed in the order that they're provided, so in this case outgoing requests will go AuthInterceptor -> EuaAuthInterceptor -> SigninAuthInterceptor (although we don't do anything to them on their way out) and incoming responses will handled SigninAuthInterceptor -> EuaAuthInterceptor -> AuthInterceptor.

I've verified that this is the behavior, and tested it with additional interceptors in an application module, where it appears to just append the additional interceptors to the list found here in the core. If interceptors are applied in several different modules, it may take some fanciness to set the order correctly.

jrassa:
I'm testing this now and it seems like the bypass logic isn't actually needed.  I am not seeing the catchError method being called on the subsequent interceptors once `router.navigate` is called.  Is there a use case you have in mind where you would want an interceptor to handle an error and bypass the remaining interceptors, but not call `router.navigate`?

sshah-asymmetrik:
Actually, now that they're not all directly related to Auth, I should change the base AuthInterceptor to just be 'Interceptor'

- [x] Rename AuthInterceptor

- [x] Make sure base interceptor isn't defaulting

sshah-asymmetrik:
@jrassa ~~You're correct, in that subsequent interceptors shouldn't get called once we navigate somewhere else. I don't need to call the `bypassRemainingInterceptor` method in the eua or signin interceptors (I'll take those calls out).~~ I'm mistaken, as I'm actually not seeing the behavior you're describing. If I have our server return a 401 to trigger the signin interceptor and put a console.log in the eua handleError method, it does in deed get called on a 401. 

The use case that I'm trying to address by including bypass logic is to allow us to not navigate anywhere at all and pass errors through to the angular components to be handled by them. (we have a 403 that we want handled by the component instead of the interceptor), so we've added a custom interceptor in our application's module that triggers the bypass to skip the default interceptors in the starter, which would otherwise redirect on a 403 no matter what.

src/app/core/auth/auth.interceptor.ts
```diff
@@ -1,44 +1,17 @@
...
-
-				return throwError(err);
-			})
-		);
+export class AuthInterceptor extends Interceptor {
+	handleError(err: any): void {
+		const { status, type, message, url } = this.parseError(err);
+		if (status === 403) {
+			this.router.navigate(['/access'], { state: { status, type, message, url } });
+			this.bypassRemainingInterceptors(err);
```
sshah-asymmetrik: This call to the bypass doesn't do anything as the AuthInterceptor is the last one in the chain, but I figured to leave it in a) for standardization and b) as a safeguard in case the order ever changes in the future.

jrassa:
> @jrassa ~You're correct, in that subsequent interceptors shouldn't get called once we navigate somewhere else. I don't need to call the `bypassRemainingInterceptor` method in the eua or signin interceptors (I'll take those calls out).~ I'm mistaken, as I'm actually not seeing the behavior you're describing. If I have our server return a 401 to trigger the signin interceptor and put a console.log in the eua handleError method, it does in deed get called on a 401.

I retested with your updates and I am still seeing the same behavior.

Here is my modified base interceptor:
```
intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
		console.log(`${this.constructor.name}: intercept`);
		return next.handle(req).pipe(
			catchError(err => {
				console.log(`${this.constructor.name}: catchError`);
				if (!err.bypassAuthInterceptors) {
					this.handleError(err);
				}
				return throwError(err);
			})
		);
	}
```

If I trigger a 403 error, I see the follow logged in the console:
```
AuthInterceptor: intercept
EuaAuthInterceptor: intercept
SigninAuthInterceptor: intercept
SigninAuthInterceptor: catchError
EuaAuthInterceptor: catchError
AuthInterceptor: catchError
```

If I get a 401 error, I see the follow logged in the console:
```
AuthInterceptor: intercept
EuaAuthInterceptor: intercept
SigninAuthInterceptor: intercept
SigninAuthInterceptor: catchError
```

jrassa:
> The use case that I'm trying to address by including bypass logic is to allow us to not navigate anywhere at all and pass errors through to the angular components to be handled by them. (we have a 403 that we want handled by the component instead of the interceptor), so we've added a custom interceptor in our application's module that triggers the bypass to skip the default interceptors in the starter, which would otherwise redirect on a 403 no matter what.

Given that use case, it sounds like you currently only need to bypass the AuthInterceptor (which handles 403 errors).  An alternate approach that I have seen is to have the interceptor check for a specific header on the request that will trigger the bypass.  Typically this flag is set at the request level.

https://medium.com/mikes-fullstack-ladder/how-to-skip-http-interceptor-dee136e54a5f

sshah-asymmetrik:
> > The use case that I'm trying to address by including bypass logic is to allow us to not navigate anywhere at all and pass errors through to the angular components to be handled by them. (we have a 403 that we want handled by the component instead of the interceptor), so we've added a custom interceptor in our application's module that triggers the bypass to skip the default interceptors in the starter, which would otherwise redirect on a 403 no matter what.
> 
> Given that use case, it sounds like you currently only need to bypass the AuthInterceptor (which handles 403 errors). An alternate approach that I have seen is to have the interceptor check for a specific header on the request that will trigger the bypass. Typically this flag is set at the request level.
> 
> https://medium.com/mikes-fullstack-ladder/how-to-skip-http-interceptor-dee136e54a5f

Ooooh good call on the header flags! Yes, in this case we really just need to bypass the AuthInterceptor. I can add a condition to the interceptors to check for a bypass header.

I'm wondering then if I should toss most of my changes to structure here. I could revert to the original AuthInterceptor (which includes the eua and signin checks) and to keep all that functionality together and avoid unnecessary complication. What do you think?

sshah-asymmetrik:
> > @jrassa ~You're correct, in that subsequent interceptors shouldn't get called once we navigate somewhere else. I don't need to call the `bypassRemainingInterceptor` method in the eua or signin interceptors (I'll take those calls out).~ I'm mistaken, as I'm actually not seeing the behavior you're describing. If I have our server return a 401 to trigger the signin interceptor and put a console.log in the eua handleError method, it does in deed get called on a 401.
> 
> I retested with your updates and I am still seeing the same behavior.
> 
> Here is my modified base interceptor:
> 
> ```
> intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
> 		console.log(`${this.constructor.name}: intercept`);
> 		return next.handle(req).pipe(
> 			catchError(err => {
> 				console.log(`${this.constructor.name}: catchError`);
> 				if (!err.bypassAuthInterceptors) {
> 					this.handleError(err);
> 				}
> 				return throwError(err);
> 			})
> 		);
> 	}
> ```
> 
> If I trigger a 403 error, I see the follow logged in the console:
> 
> ```
> AuthInterceptor: intercept
> EuaAuthInterceptor: intercept
> SigninAuthInterceptor: intercept
> SigninAuthInterceptor: catchError
> EuaAuthInterceptor: catchError
> AuthInterceptor: catchError
> ```
> 
> If I get a 401 error, I see the follow logged in the console:
> 
> ```
> AuthInterceptor: intercept
> EuaAuthInterceptor: intercept
> SigninAuthInterceptor: intercept
> SigninAuthInterceptor: catchError
> ```

This is odd. With those same logging statements I am getting the full set of 'catch error' messages when I test this in our application. I'm wondering if there are other forces at work. It may not really matter though if we switch over to the request header method. Let me know what you think about the structural changes to the interceptors (keep them or revert to the original single AuthInterceptor) and I'll make it happen!

jrassa:
I think the structural changes are good.  Can we give the base class a more descriptive name than `Interceptor`?

sshah-asymmetrik:
> I think the structural changes are good. Can we give the base class a more descriptive name than `Interceptor`?

Yes, for sure. I got a little stuck on that but for now I've just called it `AbstractHttpInterceptor`. Open to suggestions as well! I think this name is fine, but it does make things a bit more complicated as we have `HttpInterceptor` implemented by `AbstractHttpInterceptor` extended by the actual interceptors.

I've removed the 'bypass-all' method and just added a specific header check to the `AuthInterceptor`.

src/app/core/auth/abstract.interceptor.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Abstract HTTP Interceptor
+ */
+@Injectable()
+export abstract class AbstractHttpInterceptor implements HttpInterceptor {
+	protected constructor(public router: Router) {}
+
+	intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+		return next.handle(req).pipe(
+			catchError(err => {
+				if (!err.bypassAuthInterceptors) {
```
jrassa: Is this check still needed?

sshah-asymmetrik: Nope! Good catch. I've removed it, as now every implementation of this abstract interceptor should call the handleError method, which is where any bypassing will be handled via the headers.

PR #165, opened 2020-10-19
SearchInputComponent - supporting a minimum character count and searching on input events

nrwinner:
This PR refactors and extends the SearchInputComponent in several ways:
- refactor the component away from using `setTimeout` and towards using debounced RxJS subjects
- add a minimum character count and a warning that will show beneath the input in the event the search value is between (noninclusive) 0 and the min
- add ability for component to optionally search on `input` events on search input element, rather than the existing `keyup` event handlers - this way, hitting no-op keys like the ctrl/cmd, shift, etc don't trigger redundant searches

### DIFF
--- a/src/app/common/search-input/search-input.component.html
+++ b/src/app/common/search-input/search-input.component.html
@@ -1,5 +1,15 @@
-<div class="search-input-container d-flex">
-	<input [placeholder]="placeholder" type="text" class="form-control" [(ngModel)]="search" (keyup)="onKeyup()">
+<div class="search-input-container d-flex" [style.width]="width">
+	<input
+		type="text"
+		[placeholder]="placeholder"
+		class="form-control"
+		[(ngModel)]="search"
+		(keyup)="onKeyup()"
+		(input)="onInput()"
+	/>
 	<span *ngIf="search.length === 0" class="icon fa fa-search"></span>
 	<span *ngIf="search.length > 0" (click)="clearSearch($event)" class="icon fa fa-times"></span>
+</div>
+<div *ngIf="showMinCountMessage" [@minCount] class="min-count">
+	<div>Searches require a minimum of {{ minSearchCharacterCount }} characters.</div>
 </div>
--- a/src/app/common/search-input/search-input.component.html
+++ b/src/app/common/search-input/search-input.component.html
@@ -10,6 +10,6 @@
 	<span *ngIf="search.length === 0" class="icon fa fa-search"></span>
 	<span *ngIf="search.length > 0" (click)="clearSearch($event)" class="icon fa fa-times"></span>
 </div>
-<div *ngIf="showMinCountMessage" [@minCount] class="min-count">
+<div *ngIf="!disableMinCountMessage && showMinCountMessage" [@minCount] class="text-muted pt-2">
 	<div class="pt-2">Searches require a minimum of {{ minSearchCharacterCount }} characters.</div>
 </div>
--- a/src/app/common/search-input/search-input.component.html
+++ b/src/app/common/search-input/search-input.component.html
@@ -11,5 +11,5 @@
 	<span *ngIf="search.length > 0" (click)="clearSearch($event)" class="icon fa fa-times"></span>
 </div>
 <div *ngIf="showMinCountMessage" [@minCount] class="min-count">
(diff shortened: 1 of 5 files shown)


tzfx:
@nrwinner Test coverage?

nrwinner:
> @nrwinner Test coverage?

@tzfx Shiny new test suite added! ✨

tzfx (approved):
Thanks for the very nice test suite!

src/app/common/search-input/search-input.component.scss
```diff
@@ -10,3 +12,12 @@
 	float: right;
 	z-index: 500;
 }
+
+.min-count {
+	color: map-get($palette-gray, 700);
+	font-size: 14px;
```
brianghig: @nrwinner let's favor `rem` values here instead of pixels

src/app/common/search-input/search-input.component.scss
```diff
@@ -10,3 +12,12 @@
...
+
+	div {
+		padding-top: 5px;
```
brianghig: Should this be using the `pt-5` class in the template instead? We have a few different ways of providing these spacers.

src/app/common/search-input/search-input.component.ts
```diff
@@ -1,24 +1,69 @@
...
+				animate('200ms ease', style({ opacity: 0, height: 0 }))
+			])
+		])
+	]
 })
 export class SearchInputComponent {
 	@Input() placeholder = 'Search...';
 	@Output() readonly applySearch: EventEmitter<string> = new EventEmitter();
 	@Input() search = '';
+	@Input() width = '350px';
```
brianghig: @nrwinner should we be using null or `auto` as a default instead? Seems `350px` is very specific.

brianghig (changes requested):

src/app/common/search-input/search-input.component.html
```diff
@@ -1,5 +1,15 @@
...
+		[placeholder]="placeholder"
+		class="form-control"
+		[(ngModel)]="search"
+		(keyup)="onKeyup()"
+		(input)="onInput()"
+	/>
 	<span *ngIf="search.length === 0" class="icon fa fa-search"></span>
 	<span *ngIf="search.length > 0" (click)="clearSearch($event)" class="icon fa fa-times"></span>
 </div>
+<div *ngIf="showMinCountMessage" [@minCount] class="min-count">
```
jrassa: Can we use the [bootstrap](https://getbootstrap.com/docs/4.5/components/forms/) recommended styling for this?

Would simplify to the following and not require any additional css.
```
<small *ngIf="showMinCountMessage" [@minCount] class="text-muted pt-2">
	Searches require a minimum of {{ minSearchCharacterCount }} characters.
</small>
```

jrassa: Could we also have an option not show the min count message?  I did a quick test enabling it on the admin pages and it is causing the size of nearby buttons to change.

Kapture 2020-10-23 at 11 09 19

jrassa:
> * add ability for component to optionally search on `input` events on search input element, rather than the existing `keyup` event handlers - this way, hitting no-op keys like the ctrl/cmd, shift, etc don't trigger redundant searches

Any reason we shouldn't just switch to `input` or use it by default?

nrwinner:
> > * add ability for component to optionally search on `input` events on search input element, rather than the existing `keyup` event handlers - this way, hitting no-op keys like the ctrl/cmd, shift, etc don't trigger redundant searches
> 
> Any reason we shouldn't just switch to `input` or use it by default?

@jrassa I was trying to avoid a potentially breaking change for anyone relying on/handling the current keyup behavior, but from a usability standpoint I think the input event makes sense to use by default.

jrassa:
> @jrassa I was trying to avoid a potentially breaking change for anyone relying on/handling the current keyup behavior, but from a usability standpoint I think the input event makes sense to use by default.

I don't think we need to worry about that, any downstream applications should just be using the `applySearch` event emitter.

src/app/common/search-input/search-input.component.html
brianghig: @jrassa do you mean that you want to still enforce the minimum character count before emitting the search value, but simply not show the message?

brianghig (approved):

PR #170, opened 2020-11-04
Fixed navbar tooltips

rgfongheiser:
Fixed tooltips so that they show when the navbar is collapsed. Added tooltips to the lower navbar items as well.


brianghig (approved):

jrassa:
Tooltips added to menu options that open popup menus appear to be inheriting the width of the popup menu.  This differs from the link menu options that are sized based on their content.

jrassa:
There is also some conflicting behavior between the tooltip and the popup menu where the tooltip is rendered behind/over the popup menu.
Kapture 2020-11-09 at 14 57 01

jrassa:
The size issue can be solved by updating the 2 css definitions in `site-navbar.component.scss`

Current:
```css
::ng-deep .site-nav-popover-container {
...
}
::ng-deep .site-nav-popover-container {
...
}
```

Suggested Change:
```css
::ng-deep .popover.site-nav-popover-container {
...
}
::ng-deep .popover.site-nav-popover-container {
...
}
```

This also partially addresses the issue with the toolip and popover overlap.  This change ensures that the popover is always rendered above the tooltip.

PR #180, opened 2020-12-30
:bug: BLOC-704 Strip Query Parameters After Parsing

ltherrien:
**Changes**
* Moved URL splitting after query parameter parsing

### DIFF
--- a/src/app/core/auth/session.service.ts
+++ b/src/app/core/auth/session.service.ts
@@ -108,7 +108,6 @@ export class SessionService {
 		if (null == url) {
 			url = '/';
 		}
-		url = url.split('?')[0];
 
 		const queryParams = this.parseQueryParams(url);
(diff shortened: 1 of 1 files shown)


ltherrien:
All, I'm not sure of a good way to verify this besides...
* Setting your local server to PKI mode
* Opening up your application and navigating to a long URL with query parameters
* Opening up a new tab
* Pasting the URL in

This change makes sure that the query parameters are not lost on authentication.

reblace:
I only did a code review, the change looks reasonable.

I spent a while thinking about and digging into this. I can't tell when this bug was originally introduced, or if it's just always been this way. It looks like this code was significantly refactored when we went from the old mean2starter to ngx-starter, so it's possible it was introduced then. It's odd, though, since the code will clearly not work as intended. It was probably my fault :)

brianghig (approved):

PR #186, opened 2021-03-01
feat: Add support for "some" roles in AuthGuard

AedanPettit:
Added support for "some" roles in AuthGuard


src/app/core/auth/auth.guard.ts
```diff
@@ -94,15 +94,22 @@ export class AuthGuard implements CanActivate {
 
 			// compile a list of roles that are missing
 			const requiredRoles = route?.data?.roles ?? ['user'];
+			const requireAllRoles = route?.data?.requireAllRoles;
```
jrassa: For backward compatibility, requireAllRoles should default to true if not defined in route data.

jrassa (changes requested):

jrassa (approved):

PR #187, opened 2021-03-03
fix(AuthGuard): auth guard support some roles

AedanPettit:
Fixed issue from previous PR where page froze when there were no required roles


jrassa (approved):

brianghig (approved):
Code change looks good to me. @AedanPettit @jrassa do we have any test coverage for this functionality?

jrassa:
@brianghig I will submit a follow up PR adding test coverage for auth.guard

PR #195, opened 2021-04-01
Feat/modal update

sshah-asymmetrik:
With enough custom modals being created with different functionality not covered by the existing `modal.component.ts`. This PR introduces a way to 'modalize' any component and easily have it match the application's modal styling/behavior. These changes are meant to be in addition to the existing functionality, not instead of, so the existing functionality should be mostly unaffected.

**Change Summary:**
- The new `ContainerModalComponent` allows users to pass in components extending the new `AbstractModalizedDirective` to 'modalize' components as needed
- All modals will 'trap' focus within themselves, so that users tabbing through a modal will not accidentally focus background elements


src/app/common/dom-utils.service.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+		});
+		it('should return only the default focusable elements for the indicated subset of the document', () => {
+			expect(DOMUtils.getFocusableElements(document, '#subset').length).toBe(5);
+		});
+		it('should return only the supplied focusable elements', () => {
+			expect(DOMUtils.getFocusableElements(document, '#subset', 'button').length).toBe(1);
+		});
+	});
+
+	describe('#trapFocus', () => {
```
sshah-asymmetrik: - [x] Test that focus trapping works

src/app/common/modal/container-modal/container-modal.component.ts
```diff
@@ -0,0 +1,84 @@
...
+		// Add the supplied modalized component to the content container
+		this.modalizedComponentContainer.clear();
+		const componentFactory = this.componentFactoryResolver.resolveComponentFactory(
+			this.modalizedComponent
+		);
+		const componentRef = this.modalizedComponentContainer.createComponent(componentFactory);
+
+		// Set any supplied component properties
+		Object.entries(this.modalizedComponentProperties).forEach(([key, value]) => {
+			Object.defineProperty(componentRef.instance, key, {
```
sshah-asymmetrik, line 56: I had initially just been doing these assignments with `componentRef.instance[key] = value`, but that was throwing a linting error about being unable to index Type AbstractModalizableDirective by string. I believe this accomplishes the same thing, although in a less usual manner

jrassa:
For the trap focus logic, did you look at consider using the Angular CDK?  Specifically the [cdkTrapFocus directive](https://material.angular.io/cdk/a11y/overview#focustrap).

jrassa:
It would be helpful to update at least one of the existing modals as an example of how this would work in practice.

sshah-asymmetrik:
> It would be helpful to update at least one of the existing modals as an example of how this would work in practice.

Oh of course! I developed this in our project repo and did something with a project-specific modal. I'll convert one of the starters custom modals to use the new functionality and add it to the PR.

- [x] Add example of `ContainerModalComponent` in use

sshah-asymmetrik:
> For the trap focus logic, did you look at consider using the Angular CDK? Specifically the [cdkTrapFocus directive](https://material.angular.io/cdk/a11y/overview#focustrap).

I was not aware of this! I'll take a look. It seems like I may have reinvented the wheel a bit. I'll see if I can get the `cdkTrapFocus` integrated and we can cut out some code. Thanks for the tip!

- [x] Look into `cdkTrapFocus`

sshah-asymmetrik:
@jrassa I've added an example with the `site-navbar.component.ts` showing the `feedback-modal.component` which I've applied the modalizer directive to. I've also swapped out the `trapFocus` method in the `DOMUtils` with usage of `cdkTrapFocus`. It works great!

I played around with `cdkFocusInitial`, but I wasn't having much luck getting it to do what I wanted. Do you see a way to use it instead of my if statement that uses `DOMUtils.getFocusableElements`? It would be nice to not need to `DOMUtils` at all. I figured one other alternative would be to create a `focusElement` Directive that we can tack on to the first element of Modalized components. Thoughts?

One other thing that I was contemplating doing was adding a way to embed/default the `title`, `okText`, and `cancelText` in modalized components so that there are defaults within the component that the developer can choose to override via the configuration if they choose. Think it's worth implementing?

Thanks!

sshah-asymmetrik:
- [x] Fix `afterAll` build failures

sshah-asymmetrik:
@jrassa 
> I played around with `cdkFocusInitial`, but I wasn't having much luck getting it to do what I wanted. Do you see a way to use it instead of my if statement that uses `DOMUtils.getFocusableElements`? It would be nice to not need to `DOMUtils` at all. I figured one other alternative would be to create a `focusElement` Directive that we can tack on to the first element of Modalized components. Thoughts?

Nevermind on this point! Just got it. I just needed an additional attribute on the div I'm trapping the focus in.

brianghig (changes requested):
@sshah-asymmetrik it looks like this doesn't quite fit the modal spec because it still forces the user to scroll the length of the modal in order to get to the footer toolbar. From what I understood, the goal was to scroll within the modal and leave the header + footer visible at all times.

Also, it seems that the "close" button is focused by default. Is the intention to focus on the first focusable element within the contained component?

brianghig:
@sshah-asymmetrik I just added a button for the component-based modal into the "modal example" page. Hopefully that makes it easy to reproduce the scrolling issue too. Feel free to modify that example as-needed.

sshah-asymmetrik:
@brianghig 
> Also, it seems that the "close" button is focused by default. Is the intention to focus on the first focusable element within the contained component?

Oh, nope this was unintentional. I've added some description to the AbstractModalizeableDirective to indicate to developers to use the `cdkFocusInitial` attribute. Previously, the `container-modal` component would always have 'autofocus' turned on, and if there wasn't any element specified as the one to focus, it would just focus the first one it found, which would be the close button. The behavior has been updated so that now, if there's an element with `cdkFocusInitial` present, we'll turn on autofocus, but we'll leave it off otherwise.

sshah-asymmetrik:
> @sshah-asymmetrik it looks like this doesn't quite fit the modal spec because it still forces the user to scroll the length of the modal in order to get to the footer toolbar. From what I understood, the goal was to scroll within the modal and leave the header + footer visible at all times.

Added some .css to fix the position of the header/footer and have the inner content scroll. The example modal should work as expected now.

brianghig (approved):
@sshah-asymmetrik I pushed a minor update to add a top shadow to the modal footer, in order to match the style guide and provide some contrast between the footer bar and the contained component. I also decreased the size of the modal's height because it was rendering below the app's bottom banners when there were 2 stacked. At this point, this LGTM!

Will wait for @jrassa for a second set of eyes on the code + implementation before merging.

src/app/common/modal/container-modal/container-modal.component.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../../../styles/ngx-bootstrap/modal';
```
jrassa: This should not be imported here.  It will result in duplicate style definitions in the final build.

brianghig: Thanks @jrassa ! Is import the `styles/shared` the right way to go?

jrassa: That works.  Generally, in the component styles, we only want to import files that define variables/functions/mixins to avoid duplicated styles.

src/app/common/modal/container-modal/container-modal.component.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../../../styles/shared';
+
+$box-shadow-color: rgba($ux-color-modal-header, 0.5);
+
+:host,
+div[cdkTrapFocus] {
+	display: contents;
+}
```
jrassa, line 8: Bootstrap already supports the scrolling behavior we wanted, but extra elements were preventing the styles from being applied properly.  This fixes that.

src/app/common/modal/container-modal/container-modal.component.html
```diff
@@ -0,0 +1,24 @@
+<div role="dialog" cdkTrapFocus [cdkTrapFocusAutoCapture]="isCdkFocusInitial">
```
jrassa: Do we need `role="dialog"` here?  If you inspect the DOM, it is automatically added to the `<modal-container>` that ngx-bootstrap creates.

brianghig: Confirmed that we don't and removed this `role="dialog"`

src/app/common/modal/container-modal/container-modal.component.scss
brianghig: I reduced the scope of this import further from `styles/shared` to `styles/colors`, since those were the only variables we needed here. Thanks again @jrassa  !

jrassa (approved):

Issue #196, opened 2021-04-02
Create `asy-modal` content wrapper


sshah-asymmetrik:
Create a wrapper component that contains the standard pieces of a modal and allows us to insert modal content into it.

As per @jrassa's design, it's usage would look something like this:
```
<asy-modal title="Give Feedback" okText="Submit" cancelText="Cancel" (ok)="onOk()" (cancel)="onCancel" [disableOk]="!submitFeedbackForm.form.valid">
  <!-- modal body content goes here -->
  <form>
  </form>
</asy-modal>
```

This should be used in both the `modal.component` and `contianer-modal` component, which currently both contain redundant header/footer content.

jrassa:
I'm going to work on this.

tzfx:
Looks like this is done? Closing out the issue! @jrassa

PR #198, opened 2021-04-06
Add modal component for building modals

jrassa:
Adds a base modal component to use when building modals.  Abstracts away all the boilerplate template code of setting up the modals header, body, and footer.  Usage will ensure consistent styling across modals.

Address #196


src/app/common/modal/modal/modal.component.ts
```diff
@@ -1,29 +1,41 @@
...
 @Component({
-	templateUrl: 'modal.component.html'
+	selector: 'asy-modal',
+	templateUrl: './modal.component.html',
+	styleUrls: ['./modal.component.scss'],
+	changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class ModalComponent extends AbstractModalDirective {
-	message: string;
+export class ModalComponent {
```
sshah-asymmetrik, line 9: @jrassa Given the amount of overlap between these inputs and the `abstract-modal.directive` properties, would it make sense to move the inputs into the abstract directive and have the `ModalComponent` extend `AbstractModalDirective`  with the added `@Output()`s?

In the `abstract-modal.directive` We could add `@Input()` decorators to `title`, `okText`, and `cancelText` fields, and add `disableOk` and `autoCaptureFocus` inputs. Those last two are used by the `ContainerModalComponent` too, although in that component they're called `isDisableOk` and `isCdkFocusInitial`, but that's easy to change.

jrassa: I debated about extending from `AbstractModalDirective`.  Ultimately I decided against because while there is some overlap there is also enough differences, that I worry would clutter the class and potentially cause confusion (i.e. ok/cancel event emitters vs. ok/cancel methods and onClose subject) over how to use.  As well, this modal class is just focused on eliminating duplicate markup and is decoupled from the actual modal implementation (ngx-bootstrap).

If there was duplicated logic, I would be more inclined to try and consolidate, but I don't think 3 properties is enough overlap to be concerned about.

sshah-asymmetrik (approved):
Looks good to me!

PR #200, opened 2021-04-30
fix: Only include sort values in paging object if defined.

jrassa:
Currently, if sortField or sortDir are not defined, `toObj` will set them to null.  When the object gets converted to query parameters, they are included as the string 'null'.  This could cause issues on the backend.  Updating to only include if a value is set.

### DIFF
--- a/src/app/common/paging/paging.model.ts
+++ b/src/app/common/paging/paging.model.ts
@@ -59,7 +59,7 @@ export class PagingOptions {
 			page: this.pageNumber,
 			size: this.pageSize,
 			...(this.sortField && { sort: this.sortField }),
-			...(this.sortDir && { sort: this.sortDir })
+			...(this.sortDir && { dir: this.sortDir })
 		};
 	}
 }
--- a/src/app/common/paging/paging.model.ts
+++ b/src/app/common/paging/paging.model.ts
@@ -58,8 +58,8 @@ export class PagingOptions {
 		return {
 			page: this.pageNumber,
 			size: this.pageSize,
-			sort: this.sortField || null,
-			dir: this.sortDir || null
+			...(this.sortField && { sort: this.sortField }),
+			...(this.sortDir && { sort: this.sortDir })
 		};
 	}
 }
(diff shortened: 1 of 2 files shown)


src/app/common/paging/paging.model.ts
```diff
@@ -58,8 +58,8 @@ export class PagingOptions {
 		return {
 			page: this.pageNumber,
 			size: this.pageSize,
-			sort: this.sortField || null,
-			dir: this.sortDir || null
+			...(this.sortField && { sort: this.sortField }),
+			...(this.sortDir && { sort: this.sortDir })
```
tzfx: Should this be `{ dir: this.sortDir }`?

tzfx: Also, we could potentially better type PagingOptions as as `PagingOptions<T = any>`  in line with the PagingResults object, which would allow us to actually type the sortField as `(keyof T | keyof T[])`! I don't know what sort of refactoring that may cause, but it would give us some nice type checking against these objects.

jrassa: > Should this be `{ dir: this.sortDir }`?
Fixed.  Good catch.

> Also, we could potentially better type PagingOptions as as `PagingOptions<T = any>` in line with the PagingResults object, which would allow us to actually type the sortField as `(keyof T | keyof T[])`! I don't know what sort of refactoring that may cause, but it would give us some nice type checking against these objects.

One potential issue is with nested objects.  Given a schema like the following:
```js
{
  prop1: 'value1',
  prop2: {
    nestedProp1: 'value2'
  }
}
```

If you wanted to sort by `nestedProp1`, then sort field would be `prop2.nestedProp1`, which wouldn't be picked up by `(keyof T | keyof T[])`.

Another potential issue is when you want to sort by an alternative field that is not actually in the model.  An example, would be if you have a field `name`, but in the database you also have a field `name_lowercase`.  There is no need to actually return `name_lowercase` to the frontend or declare it in the model, but when sorting by Name, you really want to sort by `name_lowercase`.

tzfx: @jrassa Just ruin my day like that! :P  I forgot about nested props.

tzfx (approved):

PR #205, opened 2021-05-26
Style System Alerts

zeevo:
Many UX elements in the app are styled through variables defined in `_colors.scss`. System Alerts should be too. Currently, they are not getting styled and are untouchable `ngx-boostrap/alerts` (see https://valor-software.com/ngx-bootstrap/#/alerts#basic). This aims to fix that.

Also included are

- Add System Alerts to the example app for visibility
- Center dismissible button vertically

If this is something that looks good, I suggest we go further and...

- Add support for more `scss` variables in `_colors.scss` such as `$ux-color-alert-success-text`, `$ux-color-alert-success-border`, `$ux-color-alert-success-icon`, etc in order to allow for more customization
- Add support for an alert of type `info` to align with bootstrap's alerts. (Currently we are only styling success, warning, and danger).

Note: This may effect teams that are not aware of how they are using the `SystemAlertService`, as it will now use values in `_colors.scss` and not the default bootstrap colors.

### DIFF
--- a/src/app/site/example/alerts/alerts.component.html
+++ b/src/app/site/example/alerts/alerts.component.html
@@ -1,2 +1,30 @@
-<h1>System Alerts</h1>
-<system-alert></system-alert>
+<div class="container">
+	<h1>System Alerts</h1>
+	<div class="row">
+		<div class="col-md-3">
+			<button class="mr-3 btn btn-primary" (click)="addAlert('Success', 'success')">
+				Success
+			</button>
+		</div>
+		<div class="col-md-3">
+			<button class="mr-3 btn btn-primary" (click)="addAlert('Danger', 'danger')">
+				Danger
+			</button>
+		</div>
+		<div class="col-md-3">
+			<button class="mr-3 btn btn-primary" (click)="addAlert('Warning', 'warning')">
+				Warning
+			</button>
+		</div>
+		<div class="col-md-3">
+			<button class="mr-3 btn btn-primary" (click)="addAlert('Info', 'info')">
+				Info
+			</button>
+		</div>
+	</div>
+	<div class="row mt-3">
+		<div class="col-md-12">
+			<system-alert></system-alert>
+		</div>
+	</div>
+</div>
--- a/src/app/site/example/alerts/alerts.component.html
+++ b/src/app/site/example/alerts/alerts.component.html
@@ -1,0 +1,2 @@
+<h1>System Alerts</h1>
+<system-alert></system-alert>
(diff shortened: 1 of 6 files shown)


jrassa:
@zeevo `_colors.scss` is intended to be the base default set of reusable colors.  Most of the colors defined there correspond with the color palettes defined in the style guides created by UX.   However, no values should be changed in it when changing colors in an application built on the starter.  You should use `_overrides.scss` to override any scss variables defined elsewhere.  The styles in the starter were setup this way to reduce potential merge conflicts.

jrassa:
> * Add support for more `scss` variables in `_colors.scss` such as `$ux-color-alert-success-text`, `$ux-color-alert-success-border`, `$ux-color-alert-success-icon`, etc in order to allow for more customization

I am all for adding scss variables to allow for this further customization.  However, unless you are going to reference the same scss variables in other components, `_colors.scss` is not the place to do so.  I would suggest adding them to `system-alert.component.scss`.  This is how we handle similar customization for other components already (i.e. `flyout.component.scss`, `breadcrumb.component.scss`,  `site-container/_shared.scss`, `site-navbar/_shared.scss`) and the values can still be customized in `_overrides.scss`.

> * Add support for an alert of type `info` to align with bootstrap's alerts. (Currently we are only styling success, warning, and danger).

Sounds good.

src/app/site/example/alerts/alerts.component.html
```diff
@@ -0,0 +1,2 @@
+<h1>System Alerts</h1>
+<system-alert></system-alert>
```
jrassa: Could we add some buttons to the example page to add more alerts?

zeevo: I added 4 buttons, one for each alert type.

zeevo:
> @zeevo `_colors.scss` is intended to be the base default set of reusable colors. Most of the colors defined there correspond with the color palettes defined in the style guides created by UX. However, no values should be changed in it when changing colors in an application built on the starter. You should use `_overrides.scss` to override any scss variables defined elsewhere. The styles in the starter were setup this way to reduce potential merge conflicts.

If I wanted to completely change the theme of an application built on the starter, would I do that all within `_overrides.scss`, including palette changes and colors changes?

jrassa:
> > @zeevo `_colors.scss` is intended to be the base default set of reusable colors. Most of the colors defined there correspond with the color palettes defined in the style guides created by UX. However, no values should be changed in it when changing colors in an application built on the starter. You should use `_overrides.scss` to override any scss variables defined elsewhere. The styles in the starter were setup this way to reduce potential merge conflicts.
> 
> If I wanted to completely change the theme of an application built on the starter, would I do that all within `_overrides.scss`, including palette changes and colors changes?

Yes

Issue #72, opened 2019-09-05
Delete User button does nothing


brianghig:
From the Admin > Users page, clicking the trash icon button does not delete the user's account. No calls are made to the server, and no messages are logged into the browser's console.

tzfx:
@stella-masucci is going to take this one.

PR #209, opened 2021-06-01
issue #72 delete user functionality

stella-masucci:


src/app/core/admin/user-management/admin-list-users.component.ts
```diff
@@ -150,15 +166,12 @@ export class AdminListUsersComponent extends AbstractPageableDataComponent<User>
...
-
-		super.ngOnInit();
 	}
 
 	confirmDeleteUser(user: User) {
 		const id = user.userModel._id;
-		const username = user.userModel.username;
-
-		console.error('Delete User not yet implemented.');
+		this.adminUsersService.removeUser(id);
```
brianghig: @smasucci - let's subscribe to the `http.delete` observable in this method (vs. in the service), so that we can wait until the request completes before calling `this.reload()` in the `next` handler. That'll ensure that the user listing reflects the changes from the delete.

src/app/core/admin/user-management/admin-list-users.component.html
```diff
@@ -149,7 +149,10 @@ <h1 skipTo>
 									>
 								</li>
 								<li>
-									<a class="dropdown-item" (click)="confirmDeleteUser(user)"
+									<a
+										*ngIf="user.userModel.name != this.session.name"
```
brianghig: @smasucci - although it may seem odd, we should let logged in users delete their own account. In certain environments, these accounts will be auto-created from other authentication mechanisms (e.g., OAuth or PKI credentials), and we wouldn't want to prevent an admin from taking this action.

brianghig (changes requested):

src/app/core/admin/user-management/admin-list-users.component.ts
stella-masucci: got it!

src/app/core/admin/user-management/admin-list-users.component.html
stella-masucci: alright! in that case we don't need to get the session (it was only to check the name of the current user) so i'll take out all of those changes

src/app/core/admin/user-management/admin-list-users.component.ts
```diff
@@ -150,15 +155,18 @@ export class AdminListUsersComponent extends AbstractPageableDataComponent<User>
...
 	}
 
 	confirmDeleteUser(user: User) {
 		const id = user.userModel._id;
-		const username = user.userModel.username;
-
-		console.error('Delete User not yet implemented.');
+		this.adminUsersService.removeUser(id).subscribe({
+			next: data => {
+				this.reload();
```
brianghig: @stella-masucci - instead of setting up a second set of subscriptions for the route parameters ( what was moved into the `reload` method ), we should leave that in the `ngOnInit` method and simply call a `this.load$.next(true);` here. That will trigger a reload of the data with all of the current sorting, filtering, and paging functionality.

src/app/core/admin/user-management/admin-list-users.component.ts
```diff
@@ -150,15 +155,18 @@ export class AdminListUsersComponent extends AbstractPageableDataComponent<User>
...
+			},
+			error: error => {
+				console.error('There was an error!', error);
```
brianghig: @stella-masucci - since we have an alert service available in this class, we can let the users know whether there was a failure using something like `this.alertService.addClientErrorAlert(error);`

brianghig (changes requested):
Looks like a functioning solution, but we have a few small things to clean up for the observables. Let's also add some test coverage to this functionality so we can guarantee that we're only making one call to the `list users` API service after the selected user ID is delete.

brianghig (approved):

PR #210, opened 2021-06-02
Improve styling based on newer UX guidelines and modularize overrides

zeevo:
Hello

Our team wants to change how we style things slightly to be more inline with newer UX guidelines. This pull request does the following:

- Style Side Nav to make the active element have the same color as the brand background. Screen Shot 2021-06-02 at 1 07 47 PM
- Add defaults to palettes and add specific overrides files for better organization.

### DIFF
--- a/src/app/core/site-navbar/_shared.scss
+++ b/src/app/core/site-navbar/_shared.scss
@@ -17,7 +17,7 @@ $nav-element-hover-color: $ux-color-white !default;
 $nav-element-hover-background-color: $ux-color-primary-medium !default;
 
 $nav-element-active-background-color: $ux-color-primary-dark !default;
-$nav-element-active-color: $ux-color-brand-background !default;
+$nav-element-active-color: $ux-color-highlight !default;
 $nav-element-active-indicator-width: 4px !default;
 
 $nav-element-divider-margin: 10px !default;
--- a/src/app/core/site-navbar/_shared.scss
+++ b/src/app/core/site-navbar/_shared.scss
@@ -17,7 +17,7 @@ $nav-element-hover-color: $ux-color-white !default;
 $nav-element-hover-background-color: $ux-color-primary-medium !default;
 
 $nav-element-active-background-color: $ux-color-primary-dark !default;
-$nav-element-active-color: $ux-color-highlight !default;
+$nav-element-active-color: $ux-color-brand-background !default;
 $nav-element-active-indicator-width: 4px !default;
 
 $nav-element-divider-margin: 10px !default;
--- a/src/app/core/site-navbar/site-navbar.component.scss
+++ b/src/app/core/site-navbar/site-navbar.component.scss
@@ -77,6 +77,7 @@
 			&.active-link {
 				background-color: $nav-element-active-background-color;
 				border-left: $nav-element-active-indicator-width solid $nav-element-active-color;
+				color: $nav-element-active-color;
 
 				> span.fa {
 					margin-left: $nav-element-icon-margin-left - $nav-element-active-indicator-width;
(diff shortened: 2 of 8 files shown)


src/app/core/site-navbar/_shared.scss
```diff
@@ -17,7 +17,7 @@ $nav-element-hover-color: $ux-color-white !default;
 $nav-element-hover-background-color: $ux-color-primary-medium !default;
 
 $nav-element-active-background-color: $ux-color-primary-dark !default;
-$nav-element-active-color: $ux-color-highlight !default;
+$nav-element-active-color: $ux-color-brand-background !default;
```
jrassa: I believe this change should be handled as an override in your teams downstream project.
Changing the default will effect all downstream projects.

zeevo: I agree! Fixed.

jrassa:
> Add defaults to palettes

This change is fine, but In general, I would recommend avoiding using the palettes.  In my experience, they don't mesh all that well with bootstrap, which is why `_colors.scss` defines a set of `$ux-` that are actually used everywhere else.  Personally, I wouldn't be against removing the palette stuff.


> add specific overrides files for better organization.

I'm conflicted about breaking out the override files.  While, I kind of like the idea of the files importing the appropriate override file, I am concerned the idea will get out of hand and we will have an override file for every scss file in the starter.   I feel like we are trying to fix/organize something that isn't really a problem.  The override file for my project is ~65 lines total with comments and spacing for organization.  Of that, only about 20 lines could be moved to one of these new proposed override files.

In addition, one thing I really like about having a single override file is that when working in the starter, I can easily copy over my project's override file to preview the effect of changes in the starter.  Breaking out the overrides into multiple files makes this harder.

zeevo:
> > Add defaults to palettes
> 
> This change is fine, but In general, I would recommend avoiding using the palettes. In my experience, they don't mesh all that well with bootstrap, which is why `_colors.scss` defines a set of `$ux-` that are actually used everywhere else. Personally, I wouldn't be against removing the palette stuff.

I agree.

> > add specific overrides files for better organization.
> 
> I'm conflicted about breaking out the override files. While, I kind of like the idea of the files importing the appropriate override file, I am concerned the idea will get out of hand and we will have an override file for every scss file in the starter. I feel like we are trying to fix/organize something that isn't really a problem. The override file for my project is ~65 lines total with comments and spacing for organization. Of that, only about 20 lines could be moved to one of these new proposed override files.
> 
> In addition, one thing I really like about having a single override file is that when working in the starter, I can easily copy over my project's override file to preview the effect of changes in the starter. Breaking out the overrides into multiple files makes this harder.

Our team was encountering another issue involving having a singular `overrides` file: Some site-specific SCSS was importing specific `styles` SCSS directly, causing overrides to miss. A fix would be to import the whole `src/styles.scss` in every scenario where a developer wants to use, say a value in `_colors.scss`, and for our codebase this is a lot of changes.

Having an `@import './overrides/_colors.scss'` at the top of `_colors.scss` makes sure the overrides are applied, even when someone imports `_colors.scss` directly in their component's own `.scss` file.

jrassa:
> Our team was encountering another issue involving having a singular `overrides` file: Some site-specific SCSS was importing specific `styles` SCSS directly, causing overrides to miss. A fix would be to import the whole `src/styles.scss` in every scenario where a developer wants to use, say a value in `_colors.scss`, and for our codebase this is a lot of changes.
> 
> Having an `@import './overrides/_colors.scss'` at the top of `_colors.scss` makes sure the overrides are applied, even when someone imports `_colors.scss` directly in their component's own `.scss` file.

You definitely don't want to import the whole `src/styles.scss`.  That would result in duplicate styles.  However, there isn't really any harm in importing the overrides file multiple times since it only contains SCSS variables.  This is how it is handled in many of the component styles in the starter (i.e. `site-navbar.component.scss`).  The first 2 lines in many of these are 

```
@import '../../../styles/overrides';
@import '../../../styles/shared';
```

Actually, it wouldn't be a bad idea to update `shared` to import the overrides, since the 2 almost always are imported together (and probably should be if they are not).

PR #212, opened 2021-06-03
Import overrides at the top of shared

zeevo:
Hello

As per a comment by @jrassa #210 

> Actually, it wouldn't be a bad idea to update shared to import the overrides, since the 2 almost always are imported together (and probably should be if they are not).

Importing overrides at the top of shared just makes things easier for developers in their components.

### DIFF
--- a/src/app/common/admin/admin.component.scss
+++ b/src/app/common/admin/admin.component.scss
@@ -1,4 +1,3 @@
-@import '../../../styles/overrides';
 @import '../../core/site-container/shared';
 
 $admin-section-padding-y: $site-content-padding-y;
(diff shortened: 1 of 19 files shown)


tzfx:
@zeevo Would we now be able to remove all of the import '/path/to/overrides'; scss where shared is already imported? to include in styles.scss?

zeevo:
> @zeevo Would we now be able to remove all of the import '/path/to/overrides'; scss where shared is already imported?

Yes. If you want variables, import `src/styles/shared`

> to include in styles.scss?

Good catch

brianghig (approved):
LGTM. Hopefully @jrassa can verify too.

tzfx:
Let's remove these extra override imports if we don't need them.

Screen Shot 2021-06-03 at 16 22 51

tzfx:
Let's remove these extra override imports if we don't need them anymore:
Screen Shot 2021-06-03 at 16 25 23

jrassa:
Can we also update the following files to import shared vs. the just the colors scss file.

* `signin.component.scss`
* `admin-list-feedback.component.scss`

PR #213, opened 2021-06-03
Make palettes default and style the side nav

zeevo:
Hello

After some discussion in #210, I've pulled the relevant changes to this Pull Request instead.

- Make palettes default
- Make side bar active link a color based on newer UX guidelines

### DIFF
--- a/src/app/core/site-navbar/site-navbar.component.scss
+++ b/src/app/core/site-navbar/site-navbar.component.scss
@@ -77,6 +77,7 @@
 			&.active-link {
 				background-color: $nav-element-active-background-color;
 				border-left: $nav-element-active-indicator-width solid $nav-element-active-color;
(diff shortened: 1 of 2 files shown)


brianghig (approved):
LGTM

src/app/core/site-navbar/site-navbar.component.scss
```diff
@@ -77,6 +77,7 @@
 			&.active-link {
 				background-color: $nav-element-active-background-color;
 				border-left: $nav-element-active-indicator-width solid $nav-element-active-color;
+				color: $nav-element-active-color;
```
jrassa, line 80: Could we add another variable to allow setting the color separate from the border color? 

Suggest adding `$nav-element-active-border-color` to set the border color.

PR #220, opened 2021-08-02
build: angular 10 update

jrassa:
Updates the starter to Angular 10.  I was going to update to 11, but `ngx-bootstrap` library only added support for 11 in the last week or 2.   Would like to wait a bit to ensure any bugs are worked out first.

Downstream projects can rerun the migration schematics using the following command.
```
ng update @angular/core --from 9 --to 10 --migrate-only
```


src/app/common/pipes/utc-date-pipe/utc-date-utils.service.ts
```diff
@@ -8,13 +9,12 @@ export class UtcDateUtils {
 			let momentDate;
 			if (isMoment(value)) {
 				momentDate = value;
```
brianghig: @jrassa - are these changes simply not synced from the latest merge? I didn't expect to see `moment` here still.

brianghig (approved):

src/app/common/pipes/utc-date-pipe/utc-date-utils.service.ts
jrassa: Correct.  I will get this synced shortly.

tsconfig.json
```diff
@@ -3,14 +3,17 @@
   "compilerOptions": {
     "baseUrl": "./",
     "outDir": "./dist/out-tsc",
+    "forceConsistentCasingInFileNames": true,
```
tzfx, line 6: Are these options new additions to ts? We'll just need to pay attention to these changes in our downstream.

jrassa: These aren't new additions to ts, they are just newly recommended to be used by Angular.  As part of the update process, I generated a new Angular 10 project and compared the differences between our configuration and the Angular defaults.  These options can always be disabled downstream if necessary.

PR #222, opened 2021-08-04
Replace moment with luxon

AedanPettit:
Replaced uses of moment library with luxon. I think I covered them all but that should be double checked before we merge. 

Changed a couple of the test cases in the utc-date-utils service tests to add a "Z" at the end of the date string in order for luxon to read it properly.


jrassa:
`list-audit-entries.component.ts` and `ago-date.pipe.ts` are still using moment and need to be updated.

src/app/core/audit/list-audit-entries/list-audit-entries.component.ts
```diff
@@ -78,11 +78,11 @@ export class ListAuditEntriesComponent extends AbstractPageableDataComponent<any
 
 	dateRangeFilter: any;
 
-	queryStartDate: Date = utc()
-		.subtract(1, 'days')
-		.toDate();
+	queryStartDate: Date = DateTime.now()
+		.toUTC(1440)
+		.toJSDate();
```
jrassa: This adding vs. substracting.  Also recommend using luxon's minus function vs. the offset option on `toUTC`.

```ts
DateTime.now().minus({ days: 1 }).toUTC();
```

The minus option makes the intent clearer and is more correct.  If you convert both to ISO string format, the results are as follows:

utc offset:
2021-08-04T23:59:59.999-24:00

minus:
2021-08-04T23:59:59.999Z

jrassa (changes requested):

jrassa (approved):

package.json
```diff
@@ -30,11 +30,13 @@
...
     "@ng-select/ng-select": "^4.0",
     "@ngneat/until-destroy": "^7.3.2",
+    "@types/luxon": "^1.27.1",
     "bootstrap": "4.3.1",
     "core-js": "2.5",
     "font-awesome": "4.7",
     "jquery": "3.5",
     "lodash": "4.17",
+    "luxon": "^2.0.1",
     "moment": "2.29",
```
brianghig, line 40: @jrassa - is the idea to leave `moment` in place for downstream projects until we can all remove it? If not, we should probably remove it from this starter project. Thanks to you and @AedanPettit for working this effort!

jrassa: Yes, that is the idea.  I will prepare a followup PR to take care of that.  There is a little more to it than just removing the dependency.  We added a custom build configuration to exclude moment locale code to help reduce the final bundle size.  With this change, that will no longer be needed.

PR #226, opened 2021-09-07
Delete user confirmation prompt

AedanPettit:
Added a modal prompting the admin to confirm they want to delete a user before the user is permanently deleted


src/app/core/admin/user-management/admin-delete-user-modal.component.html
```diff
@@ -0,0 +1,27 @@
+<section class="modal-header">
```
jrassa: No need to create a specific component to confirm deletion.  Can use basic configurable modals provided by modal service.  See `view-team.component.ts#102` for example.

jrassa (changes requested):

PR #227, opened 2021-10-18
feat: Display error message when application configuration fails to load.

jrassa:
Update APP_INITIALIZER to properly throw error if config fails to load.  Add logic to application bootstrap to display error that occurs during application bootstrapping.

### DIFF
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -1,11 +1,11 @@
 import { waitForAsync, TestBed } from '@angular/core/testing';
 
 import { PopoverModule } from 'ngx-bootstrap/popover';
+import { of } from 'rxjs';
 import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
 import { ConfigService, CoreModule } from './core/core.module';
 import { SiteModule } from './site/site.module';
-import { of } from 'rxjs';
 
 describe('AppComponent', () => {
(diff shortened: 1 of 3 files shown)


jrassa:
The impetus of this PR is that we recently encountered a situation where Edge seems to cache the application more aggressively than Chrome.  Due to this, if the user were to attempt to access the application while the backend is unavailable, the cached version would load, but unable to load the application config.  This leaves the application in an odd state and created confusion for the users.


I did some research and there seems to be 2 different approaches to handling errors occurring with the APP_INITIALIZER.  The first is to halt initialization of the app and display an error message by modifying the default html.  The second, would be to allow initialization to complete, then implement a check to see if the config was loaded and if not, redirect to a configured route that will display an error message.

This PR is implementing the first approach, but I am not sure if it is the better/preferred approach.  @brianghig @tzfx any thoughts?

tzfx:
From an organizational standpoint like the idea of giving the downstream projects full control of what this failure mode looks like without mucking about in the core/main area.

However, if the initial application config fails, we may end up in a place where the application shouldn't be initialized at all... So, I think I'm fine with the implementation you have here. Though it looks like there may be some test failures.

tzfx (changes requested):
I'm good with this, assuming the build passes.

tzfx (approved):

PR #228, opened 2021-11-05
refactor: site-container update

jrassa:
refactors the site-container to use css flexbox layout vs. the current method which uses fixed position elements and margins to offset the main content.


jrassa:
The big advantage of this change is that scrolling is now contained to the site content area vs. the whole page.  This opens up the ability to have components that are anchored to the page viewport.

PR #231, opened 2021-11-12
fix(team): ensure user session is properly reloaded after team operations

jrassa:


src/app/core/teams/create-team/create-team.component.ts
```diff
@@ -127,7 +127,7 @@ export class CreateTeamComponent implements OnInit {
 		this.teamsService
 			.create(this.team, this?.teamAdmin?.userModel._id)
 			.pipe(
-				tap(() => this.authenticationService.reloadCurrentUser()),
+				switchMap(() => this.sessionService.reloadSession()),
```
jrassa, line 130: Just switching the call to `reloadSession` wasn't enough.  When using the `tap` operator, the returned Observable is never subscribed to, so the session is never reloaded.  Using `switchMap` ensures that the returned observable is subscribed to properly.

src/app/core/teams/team-authorization.service.ts
```diff
@@ -17,7 +17,6 @@ export class TeamAuthorizationService {
 		this.sessionService
 			.getSession()
 			.pipe(
-				first(),
```
jrassa, line 20: Even once the session was being reloaded with the fix above, the team-authorization service wasn't actually listening for the updated session.  Removing the `first` operator here ensures that the service gets the latest session.

brianghig:
@jrassa - I pushed a few more updates to ensure that the user's session is refreshed after updating team roles, deleting a team, etc.

LGTM now, but we could certainly use your eyes on before merging. In the meantime, we'll test this in our fork.

Issue #229, opened 2021-11-09
Team's page display issues


brianghig:
To improve the consistency of the information displayed on the Team details page, we should make the following updates ( or something similar ):

1. Move the "Date Created" into the "Basic Information" section and out of the header field
2. Fix the date display for the "Date Created" field
3. Use the input capitalization that the user provided vs. changing all of the "Description" input to capital case
4. Support new lines in the display of the "Description" field, both in the Team page and the Teams Listing

<img width="585" alt="Screen Shot 2021-11-09 at 5 43 12 PM" src="https://user-images.githubusercontent.com/297404/141017572-8557ab6d-982b-4d8c-9bb5-a349196fee1d.png">

<img width="550" alt="Screen Shot 2021-11-09 at 5 43 17 PM" src="https://user-images.githubusercontent.com/297404/141017588-81af0309-1a46-46f8-81d6-6ecc2249e801.png">

<img width="560" alt="Screen Shot 2021-11-09 at 5 43 24 PM" src="https://user-images.githubusercontent.com/297404/141017603-99ce1368-2964-4737-a126-a2245f8bb193.png">

jrassa:
> * Support new lines in the display of the "Description" field

Would this apply only to the display in the settings card?  Or should the list teams page be updated as well?

> * Consider moving the "Description" under the Team Name in smaller text

Would the new lines be displayed here as well?  It could be problematic with a description with a lot of new lines.  It would push the main content down.

jrassa:
I would also recommend updating the `Settings` header to be `Details`.

brianghig:
Thanks, @jrassa . I updated the description to reflect that.

In addition to renaming "Settings" to "Details", can we remove the "Basic Information" header? Are there other sections of Team information that would go here? If not, it seems like those headers are redundant.

jrassa:
> In addition to renaming "Settings" to "Details", can we remove the "Basic Information" header? Are there other sections of Team information that would go here? If not, it seems like those headers are redundant.

That is fine.  That header is there to mimic similar views in downstream app that does have multiple sections.

PR #244, opened 2022-01-20
fix: misc team fixes

jrassa:
* default implicitMembers to false when creating new team
* update required roles on auth guard for create team page
* set default filter for getTeams method
* fix invalid date format on view teams page


jrassa:
This address #229 and a few other team issues

PR #263, opened 2022-04-26
fix: AsyHeaderDateFilter generating too large of date ranges

jrassa:
When specifying a duration, the generated filter would be larger than intended by one duration.
i.e.:
1 week -> 2 weeks
3 months -> 4 months

This corrects that issue and add test coverage.

### DIFF
--- a/src/app/common/table/filter/asy-header-date-filter/asy-header-date-filter.component.ts
+++ b/src/app/common/table/filter/asy-header-date-filter/asy-header-date-filter.component.ts
@@ -92,7 +92,8 @@ export class AsyHeaderDateFilterComponent extends AsyAbstractHeaderFilterCompone
 
 		if (this.enabled) {
 			if (!this.isCustom) {
-				const now = DateTime.now();
+				const now = DateTime.utc();
+
 				if (this.direction === 'past') {
 					$gte = now.minus({ [this.duration]: this.count });
 					$lte = now;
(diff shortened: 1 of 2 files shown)


src/app/common/table/filter/asy-header-date-filter/asy-header-date-filter.component.ts
```diff
@@ -93,24 +93,24 @@ export class AsyHeaderDateFilterComponent extends AsyAbstractHeaderFilterCompone
...
 			}
-		}
-		// Normalize to start/end of day
-		if (this.isCustom || this.duration !== 'hour') {
-			$gte = $gte.startOf('day');
-			$lte = $lte.endOf('day');
+
+			// Normalize to start/end of day
+			if (this.isCustom || this.duration !== 'hour') {
+				$gte = $gte.startOf('day');
```
brianghig, line 111: @jrassa - when we're finding the start and end of the day, is that localized to the user's start and end? Or using some standardized start / end, like UTC?

jrassa: It is standardized to UTC.

jrassa: Actually, previously it was localized to the user's start/end, but will now be standardized to UTC.

brianghig: Sounds good - thanks!

brianghig (approved):
LGTM

PR #278, opened 2022-08-22
fix: Reset table paging to first page when sort/filter changes

jrassa:
Fixes bug where invalid page is requested if sort/filter results in less pages than the current page.

i.e. User is viewing page 2 of table, but then applies a sort/filter that results in only one page of results.  App currently still requests page 2, even though it has no data.

### DIFF
--- a/src/app/common/table/asy-table-data-source.ts
+++ b/src/app/common/table/asy-table-data-source.ts
@@ -1,7 +1,7 @@
 import { DataSource } from '@angular/cdk/collections';
 
 import isEmpty from 'lodash/isEmpty';
-import { combineLatest, BehaviorSubject, Observable, Subject, Subscription } from 'rxjs';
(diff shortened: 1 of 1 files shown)


src/app/common/abstract-entity.service.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export enum ServiceMethod {
+	create = 'create',
+	get = 'get',
+	update = 'update',
+	search = 'search',
+	remove = 'remove'
+}
+
+export abstract class AbstractEntityService<T extends { _id: string }> {
```
brianghig: @jrassa - was this class added intentionally? Or did you just want to update the `asy-table-data-source.ts` file?

It looks good to me overall, and a great candidate for test coverage with a concrete test service!

jrassa: I did not intend to include this class in this PR.  Cleaning it up now.

PR #279, opened 2022-08-23
refactor: Abstract CRUD+ functionality into base service class

jrassa:


jrassa:
Depends on bluehalo/node-rest-starter#191.

src/app/core/teams/teams.service.ts
```diff
@@ -21,95 +22,47 @@ export interface AddedMember {
...
-		query: any = {},
-		search: string = '',
-		options: any = {}
-	): Observable<PagingResults<Team>> {
+	override create(team: Team, firstAdmin?: string): Observable<any> {
 		return this.http
-			.post<PagingResults>(
-				'api/teams',
-				{ s: search, q: query, options },
+			.put(
```
brianghig: @jrassa - with the server-side change to make this a POST, this line needs to also stay as a POST vs. PUT.

brianghig (changes requested):

brianghig:
Changes look great to me otherwise! Nice work @jrassa 

Tested and verified UI functionality, except for two issues around the feedback flyout:
1. When submitting "Ask a Question or Comment" feedback, the classification from the dropdown is lost. The field is an empty string when returned from the UI's API.
2. During that same submission, "undefined" is prepended to the "body" attribute. This string is carrying through to the feedback record stored in the database.

src/app/core/teams/teams.service.ts
jrassa: Fixed.  Originally, I started switching the other endpoints to use PUT, then realized that teams was the only one using PUT, so switched back to POST.  Missed updating this.

jrassa:
> Changes look great to me otherwise! Nice work @jrassa
> 
> Tested and verified UI functionality, except for two issues around the feedback flyout:
> 
> 1. When submitting "Ask a Question or Comment" feedback, the classification from the dropdown is lost. The field is an empty string when returned from the UI's API.
> 2. During that same submission, "undefined" is prepended to the "body" attribute. This string is carrying through to the feedback record stored in the database.

I think this may be a configuration issue on your end.

I have the feedback classifications configured as follows and it is working properly for me.
```
feedback: {
	showFlyout: true,
	showInSidebar: true,

	classificationOpts: [
		{ level: 'LEVEL-1', prefix: '(L1)' },
		{ level: 'LEVEL-2', prefix: '(L2)' },
		{ level: 'LEVEL-3', prefix: '(L3)' }
	]
},
```

However, if I changed it to have just strings, then I see the behavior that you are describing.

```
feedback: {
	showFlyout: true,
	showInSidebar: true,

	classificationOpts: [
		'LEVEL-1',
		'LEVEL-2'
	]
},
```

I pushed a changed to node-rest-starter to updated the default feedback config.

brianghig:
> I think this may be a configuration issue on your end.

@jrassa - you're absolutely right! Changes look good to me. Thanks for this clean + consistent update.

brianghig (approved):
LGTM

Issue #280, opened 2022-09-08
Upgrade to Angular 14


brianghig:
- Target the latest release of 14.2.1
- Ensure that examples, including forms, are updated to reflect the latest Angular and ngx-starter standards
- See https://blog.angular.io/angular-v14-is-now-available-391a6db736af for an overview of changes in Angular 14

PR #282, opened 2022-09-13
build: Update to Angular 14.2

jrassa:


src/app/core/admin/user-management/admin-create-user.component.ts
```diff
@@ -18,16 +18,8 @@ import { ManageUserComponent } from './manage-user.component';
 export class AdminCreateUserComponent extends ManageUserComponent {
 	mode = 'admin-create';
 
-	constructor(
-		router: Router,
-		configService: ConfigService,
-		alertService: SystemAlertService,
-		private adminUsersService: AdminUsersService
-	) {
+	constructor(private adminUsersService: AdminUsersService) {
```
eacaps: Is there any reason this was not changed to an inject?

src/app/core/admin/user-management/admin-edit-user.component.ts
```diff
@@ -19,22 +19,8 @@ import { ManageUserComponent } from './manage-user.component';
...
-		super(
-			router,
-			configService,
-			alertService,
-			'Edit User',
-			"Make changes to the user's information",
-			'Save',
-			'/admin/users'
-		);
+	constructor(private route: ActivatedRoute, private adminUsersService: AdminUsersService) {
```
eacaps: injects?

src/app/core/auth/directives/has-every-role.directive.ts
```diff
@@ -14,12 +14,11 @@ import { SessionService } from '../session.service';
 export class HasEveryRoleDirective extends AbstractIfThenElseDirective implements OnInit {
 	roles: Array<string | Role>;
 	constructor(
-		viewContainer: ViewContainerRef,
 		templateRef: TemplateRef<any>,
 		private sessionService: SessionService,
```
eacaps, line 17: allegedly `inject` should be used for directives as well?

src/app/core/auth/directives/has-role.directive.ts
```diff
@@ -14,12 +14,11 @@ import { SessionService } from '../session.service';
 export class HasRoleDirective extends AbstractIfThenElseDirective implements OnInit {
 	role: string | Role;
 	constructor(
-		viewContainer: ViewContainerRef,
 		templateRef: TemplateRef<any>,
 		private sessionService: SessionService,
```
eacaps: inject?

src/app/core/auth/directives/has-some-roles.directive.ts
```diff
@@ -14,12 +14,11 @@ import { SessionService } from '../session.service';
 export class HasSomeRolesDirective extends AbstractIfThenElseDirective implements OnInit {
 	roles: Array<string | Role>;
 	constructor(
-		viewContainer: ViewContainerRef,
 		templateRef: TemplateRef<any>,
 		private sessionService: SessionService,
```
eacaps: inject?

src/app/core/auth/directives/is-authenticated.directive.ts
```diff
@@ -15,12 +15,11 @@ export class IsAuthenticatedDirective extends AbstractIfThenElseDirective implem
 	private _isAuthenticated = true;
 
 	constructor(
-		viewContainer: ViewContainerRef,
 		templateRef: TemplateRef<any>,
 		private sessionService: SessionService,
```
eacaps: inject?

eacaps (commented):
Minor questions, overall looks like a good upgrade

src/app/core/admin/user-management/admin-edit-user.component.ts
jrassa: The main goal I tried to achieve with using `inject` was to reduce code.  The main focus was on classes that extend a base class that had to declare dependencies in the constructor and pass them to the base class through a `super` call in the constructor. 

In this case, `route` and `adminUsersService` are not passed down to the base class, so I left it in the constructor.

src/app/core/admin/user-management/admin-create-user.component.ts
jrassa: The main goal I tried to achieve with using `inject` was to reduce code.  The main focus was on classes that extend a base class that had to declare dependencies in the constructor and pass them to the base class through a `super` call in the constructor. 

In this case, `adminUsersService` is not passed down to the base class, so I left it in the constructor.

src/app/core/auth/directives/has-role.directive.ts
jrassa: For all of these classes that extend from `AbstractIfThenElseDirective`, I choose not to use `inject` for `viewContainer`.  The reasoning was that I was unsure how it would interact in the case that the thenTemplate is overridden like in the example below, and decided to leave it as is.

```html
<div *hasRole="'admin'; then thenBlock"></div>
<ng-template #thenBlock>...</ng-template>
```

src/app/core/auth/directives/is-authenticated.directive.ts
jrassa: For all of these classes that extend from `AbstractIfThenElseDirective`, I choose not to use `inject` for `viewContainer`.  The reasoning was that I was unsure how it would interact in the case that the thenTemplate is overridden like in the example below.

```html
<div *isAuthenticated="true; then thenBlock"></div>
<ng-template #thenBlock>...</ng-template>
```

src/app/core/auth/directives/has-some-roles.directive.ts
jrassa: For all of these classes that extend from `AbstractIfThenElseDirective`, I choose not to use `inject` for `viewContainer`.  The reasoning was that I was unsure how it would interact in the case that the thenTemplate is overridden like in the example below.

```html
<div *hasSomeRoles="['user', 'admin']; then thenBlock"></div>
<ng-template #thenBlock>...</ng-template>
```

src/app/core/admin/user-management/admin-create-user.component.ts
eacaps: I prefer the usage of `inject` vs the constructor so I would just move services out of the constructor when the file is modified. Here's a quick article I found that I generally agree with: https://netbasal.com/angular-di-using-inject-instead-of-constructor-d0782d434fb1. Not going to hold up this PR over it tho

src/app/core/auth/directives/has-some-roles.directive.ts
eacaps: In line with my last comment, while making this change I would advocate for just moving `sessionService` and `authorizationService` to injects rather than in the constructor, but again this is more of a nit pick.

eacaps:
Closes #280

src/app/core/admin/user-management/admin-create-user.component.ts
jrassa: I'm convinced.  Pushing an update now.

eacaps (approved):

PR #292, opened 2022-12-22
Issue #290 - adding new config value for user deletion

### LINKED ISSUE #290: Add Configuration option for delete user
Add configuration option to disable the ability to delete a user.

Disabling user deletion should have two effects.

* Remove the Delete option from the UI on the admin user list page.
* Prevent user deletion via API.  Either remove the delete endpoint entirely or update endpoint to throw error when deletion is disabled.

stella-masucci:


src/app/core/admin/user-management/list-users/admin-list-users.component.html
```diff
@@ -180,7 +180,10 @@ <h1 skipTo>Users</h1>
 									>
 								</li>
 								<li>
-									<a class="dropdown-item" (click)="confirmDeleteUser(user)"
+									<a
```
bgdoyle: I would guess we would rather have the `*ngIf` on the `<li>` element. As it is now, there will be an empty li element in the list, which is probably not intended.

src/app/core/config.model.ts
```diff
@@ -38,6 +38,8 @@ export interface Config {
 	version: string;
 	contactEmail: string;
 
+	allowDelete: boolean;
```
bgdoyle: I think `allowDelete` as a top-level config value is too ambiguous, though I'm not sure if @jrassa has any thoughts on a better location or name for it.

src/app/core/site-navbar/site-navbar.component.spec.ts
```diff
@@ -44,7 +44,8 @@ describe('Site Navbar Component Spec', () => {
 		contactEmail: 'test@test.com',
 		teams: { implicitMembers: { strategy: 'local' } },
 		version: 'test',
-		masqueradeEnabled: false
+		masqueradeEnabled: false,
+		allowDelete: true
```
bgdoyle: @jrassa do we want a unit test to check that `allowDelete: false` actually removes the option from the dropdown?

src/app/core/config.model.ts
stella-masucci: yeah... is allowUserDelete better?

jrassa: Yes

src/app/core/site-navbar/site-navbar.component.spec.ts
jrassa: Not necessary.

jrassa (approved):

PR #293, opened 2022-12-29
Issue #160 - EUA feature

### LINKED ISSUE #160: End User Agreement (EUA) Feature
Make the current EUA available for viewing at any time by users.

Add option to the user popover menu.  Option should only be visible if there is currently a published EUA.  Clicking option should redirect user to existing `/eua` route.

Update `UserEuaComponent`.  If user has already acknowledged the EUA:
* hide acknowledgement checkbox
* hide submit button

stella-masucci:
Adds option to the user popover menu to show current EUA if one exists

### DIFF
--- a/src/app/core/eua/user-eua.component.html
+++ b/src/app/core/eua/user-eua.component.html
@@ -3,7 +3,9 @@
 		<div *ngIf="eua$ | async as eua" class="col-xl-8 offset-xl-2">
 			<h1>
 				End User Agreement (EUA)<br />
-				<small>Please review and acknowledge the end user agreement</small>
+				<small *ngIf="!alreadyAccepted"
+					>Please review and acknowledge the end user agreement</small
+				>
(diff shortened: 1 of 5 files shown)


bgdoyle:
The EUA menu item does not show up after initial sign in.
It requires a refresh of the page, most likely due to the initial `/eua` call returning a 401 because there is not a signed in user.

I'm guessing the intended behavior would be to have this show up when a user signs in?
Setting `this.eua = session?.user?.eua` inside the `getSession` subscription seems to accomplish this and doesn't fetch the eua an extra time.

Other than that, I was able to verify the behavior of the EUA routing and display page.
It showed up for a user after an EUA was created.
I then deleted the EUA as an admin, then logged back in as the normal user.
The EUA menu item was not there after deleting the EUA.

jrassa (approved):

PR #304, opened 2023-02-07
feat: Improve accessibility of navbar

jrassa:
Updated navbar links so you can now navigate to them using keyboard tabbing. This happened because <a> elements with no href attribute will be ignored, and to a user with navigating the website through keyboard alone, those links will be completely inaccessible. This commit adds a directive that adds the `tabindex="0"`, and then forces an artificial click on submit to copy native link behavior

### DIFF
--- a/src/app/common/directives/link-accessibility.directive.ts
+++ b/src/app/common/directives/link-accessibility.directive.ts
@@ -1,9 +1,10 @@
 import { Directive, ElementRef, HostListener, Renderer2 } from '@angular/core';
 
 /**
- * Used to mark where in a page the "Skip to main content" link
- * should skip to. It is ideally placed on the page's main heading:
- * the h1 element. This directive should only be used once per page
+ * Used to enable link for keyboard accessibility.
+ * This is only necessary for interactibles using that do not have a
+ * href attribute, because those will naturally be ignored by
+ * the tabbing structure.
  */
 @Directive({
 	selector: '[linkAccessibility]'
--- a/src/app/common/directives/link-accessibility.directive.ts
+++ b/src/app/common/directives/link-accessibility.directive.ts
@@ -1,0 +1,22 @@
+import { Directive, ElementRef, HostListener, Renderer2 } from '@angular/core';
+
+/**
+ * Used to mark where in a page the "Skip to main content" link
+ * should skip to. It is ideally placed on the page's main heading:
+ * the h1 element. This directive should only be used once per page
+ */
+@Directive({
+	selector: '[linkAccessibility]'
+})
+export class LinkAccessibilityDirective {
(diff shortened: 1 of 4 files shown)


src/app/common/directives/link-accessibility.directive.ts
```diff
@@ -0,0 +1,22 @@
+import { Directive, ElementRef, HostListener, Renderer2 } from '@angular/core';
+
+/**
+ * Used to mark where in a page the "Skip to main content" link
+ * should skip to. It is ideally placed on the page's main heading:
+ * the h1 element. This directive should only be used once per page
+ */
```
jrassa: Need to update or remove comment.

src/app/common/directives/link-accessibility.directive.ts
```diff
@@ -0,0 +1,22 @@
...
+@Directive({
+	selector: '[linkAccessibility]'
+})
+export class LinkAccessibilityDirective {
+	constructor(private elRef: ElementRef, private renderer: Renderer2) {
+		const el = elRef.nativeElement;
+		renderer.setAttribute(el, 'tabIndex', '0');
+	}
+
+	@HostListener('keydown.enter', ['$event'])
```
jrassa, line 18: Should also respond to pressing `space`

src/app/common/directives/link-accessibility.directive.ts
```diff
@@ -0,0 +1,22 @@
...
+	onEnter(event: Event) {
+		let targetEl = event.target;
+		targetEl?.dispatchEvent(new Event('click'));
+	}
```
jrassa: This doesn't work as intended when `<a>` has both a href/routerLink and a click event.  In this case, click event is executed, but the href is never followed.   This is the case for all of the options in the admin popup.  May be the case for others as well.

src/app/common/directives/link-accessibility.directive.ts
lu-c: Updated!

src/app/common/directives/link-accessibility.directive.ts
lu-c: Now also responds to space

src/app/common/directives/link-accessibility.directive.ts
lu-c: Added an explicit href router navigation. I don't know if this is the cleanest way of doing it, but it works!

jrassa: Still not working for any of the options in the admin menu.  Also not working for the `Api Docs` option in the help menu.

jrassa: For those, the solution might just be to remove the `linkAccessibility` directive.  Since they have hrefs defined, not sure if they need it.

PR #306, opened 2023-02-07
fix: sort headers styling now indicates what column is sorted on

lu-c:
**Before**
screenshot_2023-02-07_at_1 08 39_pm
screenshot_2023-02-07_at_1 08 44_pm

**After**
screenshot_2023-02-07_at_1 08 16_pm
screenshot_2023-02-07_at_1 10 08_pm


jrassa (approved):
This was working previously, but looks like it broke when we updated Font Awesome recently.  Good catch.

PR #335, opened 2023-08-28
refactor: Improve typings when using cdk dialog

jrassa:


jrassa:
Closing in favor of #336

